PR #12, opened 2022-12-12
Add globbing and ignore support

voxpelli:
To enable the creation of a report for a complex mono-repo proper globbing and ignore support is needed.

We eg. do not want to include the `package.json` from all the `node_modules` dependencies of a project.

This uses a default ignore list from [`ignore-by-default`](https://www.npmjs.com/package/ignore-by-default) 0 the built in [`.gitignore` support in `globby`](https://github.com/sindresorhus/globby#gitignore) as a base for ignores, then adds our own `projectIgnorePaths` from the new version of our [`socket.yml`](https://docs.socket.dev/docs/socket-yml) on top of that.


bcomnes (approved):

lib/utils/socket-config.js
```diff
@@ -0,0 +1,66 @@
...
+  },
+  required: ['version'],
+  additionalProperties: true,
+}
+
+/**
+ * @param {string} filePath
+ * @returns {Promise<SocketYml|undefined>}
+ */
+export async function readSocketConfig (filePath) {
```
bcomnes, line 42: Should break this function up into 2 steps when we modularize this. A function that parses and validates string content, and another that wraps the first function that includes the file io step.

voxpelli: Yep, already done that in my extracted version 🙂

bcomnes: 👌

PR #17, opened 2023-01-09
Add cjs-based `getCommands()` and `getHelpers()`

voxpelli:
This enables us and others to make use of some of the features directly in code rather than to go through the CLI interface

### DIFF
--- a/index.cjs
+++ b/index.cjs
@@ -1,0 +1,42 @@
+'use strict'
+
+/**
+ * @typedef SocketCliJsCommands
+ * @property {import('./lib/commands/report/create.js').createReport} createReport
+ */
+
+/**
+ * @typedef SocketCliJsHelpers
+ * @property {import('./lib/utils/path-resolve.js').getPackageFiles} getPackageFiles
+ */
+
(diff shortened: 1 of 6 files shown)


voxpelli:
Giving it another look, `createReport()` is just a thin wrapper on top of `socketSdk.createReportFromFilePaths(packagePaths, cwd)`

The better solution is probably to move the `getPackageFiles` helper to the SDK and point everyone to the SDK 🤔

bmeck:
is this missing the index.cjs output?

bcomnes (approved):

index.cjs
```diff
@@ -0,0 +1,42 @@
...
+/** @returns {Promise<SocketCliJsCommands>} */
+async function getCommands () {
+  const [
+    { createReport }
+  ] = await Promise.all([
+    import('./lib/commands/report/create.js'),
+  ])
+
+  return {
+    createReport,
```
bmeck, line 22: this needs the `import.meta` bound because `meow` needs it, so it isn't quite usable as it is.

PR #25, opened 2023-03-02
Add npm provenance GH action

bmeck:
This will create an `npm publish` whenever a release is created on GH with proper npm provenance.

### DIFF
--- a/.github/workflows/provenance.yml
+++ b/.github/workflows/provenance.yml
@@ -1,0 +1,26 @@
+name: Publish Package to npmjs
+
+on:
+ release:
+   types: [created]
+
+jobs:
+ build:
+   runs-on: ubuntu-latest
+
+   permissions:
+     contents: read
+     id-token: write
+
+   steps:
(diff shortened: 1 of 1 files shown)


voxpelli:
@bmeck Can you coordinate with @bcomnes in https://github.com/SocketDev/socket-config-js/pull/15 ?

bcomnes (approved):
Since provenance is a weird beta feature now, lets just keep this simple. This looks good for now. @voxpelli do does the CLI need any sort of build step during versioning/publishing?

voxpelli:
@bcomnes Any such step would have been in `prepublishOnly`, but I think that if we should do auto-publishing from GH Actions then we should do it the same in all places? And if we don't want to add `--provenance` to all, then we can add a flag for that?

bcomnes:
I see, the difference here is this centers around a use versioning and tagging locally, and the config does that step in the action.  I prefer the later, but starting here is fine with me. Your call.

PR #39, opened 2023-04-07
tty ipc to mitigate subshells prompts

bmeck:
this handles subshells that have tty removed by making an ipc bridge to the original pid


mikolalysenko (approved):
Let's try this out!

Random thought: have you given much consideration to what happens if one of the processes exits unexpectedly during this negotiation?  This is probably an error case anyway, but I'm wondering if it exits gracefully or does something a bit weird like auto installing everything.

mikolalysenko (approved):

PR #45, opened 2023-05-31
CLI login/logout

101arrowz:


lib/utils/settings.js
```diff
@@ -0,0 +1,56 @@
...
+
+/** @type {any} */
+let settings = {}
+
+if (fs.existsSync(settingsPath)) {
+  const raw = fs.readFileSync(settingsPath)
+  try {
+    settings = JSON.parse(ascii85.decode(raw).toString())
+  } catch (e) {
+    // failed decode - don't load
```
bmeck, line 31: this should probably just throw since it seems corrupted, and log an error

lib/utils/settings.js
```diff
@@ -0,0 +1,56 @@
+import * as fs from 'fs'
+import * as os from 'os'
+import * as path from 'path'
+
+// @ts-ignore no types for ascii85
+import ascii85 from 'ascii85'
+
+let dataHome = process.platform === 'win32'
```
bmeck, line 8: I'd say just use a lib off npm instead like https://socket.dev/npm/package/xdg-portable , that will reduce maintenance burden / complexity

lib/utils/settings.js
```diff
@@ -0,0 +1,56 @@
...
+  } catch (e) {
+    // failed decode - don't load
+  }
+} else {
+  fs.mkdirSync(path.dirname(settingsPath), { recursive: true })
+}
+
+/**
+ * @param {string} key
+ * @returns {any}
```
bmeck, line 39: This should be able to be templated for better types, same for updateSetting:

```suggestion
 * @template {string} Key
 * @param {Key} key
 * @returns {typeof settings[Key]}
```

lib/utils/settings.js
```diff
@@ -0,0 +1,56 @@
...
+  const home = os.homedir()
+  dataHome = path.join(home, ...(process.platform === 'darwin'
+    ? ['Library', 'Application Support']
+    : ['.local', 'share']
+  ))
+}
+
+const settingsPath = path.join(dataHome, 'socket', 'settings')
+
+/** @type {any} */
```
bmeck, line 23:
```suggestion
/** @type {{apiKey?: string | null} */
```

lib/commands/login/index.js
```diff
@@ -0,0 +1,61 @@
...
+
+    if (cli.input.length) cli.showHelp()
+
+    if (!isInteractive()) {
+      throw new InputError('cannot prompt for credentials in a non-interactive shell')
+    }
+    const { apiKey } = await prompts({
+      type: 'password',
+      name: 'apiKey',
+      message: 'Enter your Socket.dev API key',
```
bmeck, line 39: this should include a rough link to the basic dashboard somehow or the docs so a user could find where the API key comes from

bmeck (changes requested):
This file should be used by the npm integration and other commands

lib/utils/settings.js
101arrowz: I'm dissuaded by the low Socket score of all existing options + the fact this is only like 4 lines of code

PR #46, opened 2023-06-01
Upgraded CLI login

101arrowz:
Dependent upon #45 

These changes make a few assumptions:
- a `/v0/settings` endpoint will exist in the future
- it will return data in this format:
```js
{
  organizations: {
    orgid1: {
      name: "Org 1",
      id: "orgid1",
      plan: { tier: "enterprise" /* or "opensource" or "team" */ },
      issueRules: {
        typosquat: { action: "warn" /* or "error" or "ignore" or "defer" */ }
        malware: true // or false: true = error, false = ignore, undefined = defer
      }
    },
    // any other orgs here...
  },
  defaultIssueRules: { /* same as issueRules, but no undefined rules and no defers */ }
}
```

It does not yet provide support for `socket.yml` - that can be done in a different PR

PR #47, opened 2023-06-04
Python CLI support

101arrowz:
Depends on https://github.com/SocketDev/socket-sdk-js/pull/95. `socket pip` is WIP


lib/commands/report/create.js
```diff
@@ -181,7 +181,17 @@ async function setupCommand (name, description, argv, importMeta) {
       }
     })
 
-  const packagePaths = await getPackageFiles(cwd, cli.input, config, debugLog)
+  // TODO: setupSdk(getDefaultKey() || FREE_API_KEY)
```
bmeck, line 184: since the login is supported can we do this now?

bmeck (approved):

lib/commands/report/create.js
101arrowz: Gotta wait until #46 is in unfortunately

PR #60, opened 2023-07-10
"warn" action for `socket npm`

bmeck:


101arrowz:
This is already partially implemented in #46 - might want to make sure there are no conflicts there

101arrowz (approved):

package.json
```diff
@@ -15,7 +15,7 @@
   },
   "license": "MIT",
   "engines": {
-    "node": "^14.18.0 || ^16.13.0 || >=18.0.0"
+    "node": "^16.13.0 || >=18.0.0"
```
karfau, line 18: Are you intentionally dropping support for node 14 here? And if so, why?

bmeck: Several dependencies dropped support for it, and it is out of LTS completely, not even in maintenance https://nodejs.dev/en/about/releases/

PR #92, opened 2023-11-09
Handle dist tags / no version argument when running the info command

charliegerard:
## PR description

At the moment, users can't run the command `socket info` without specifying a package version. For example, running `socket info webtorrent` returns the error `✖ Invalid input: Need to specify a full package identifier, like eg: webtorrent@1.0.0`, event hough our help command indicates it should work.

```
  Examples
    $ socket info webtorrent
    $ socket info webtorrent@1.9.1
```
To improve the user experience, I updated the code so that the command can be run without specifying a version or also with dist tags such as `latest`, `beta`, etc.

Example: `socket info lodash` or `socket info lodash@latest`

Screenshot 2023-11-28 at 2 09 06 PM

### DIFF
--- a/lib/commands/info/index.js
+++ b/lib/commands/info/index.js
@@ -2,6 +2,7 @@
 
 import chalk from 'chalk'
 import meow from 'meow'
+import fetch from 'node-fetch'
 import ora from 'ora'
 
 import { outputFlags, validationFlags } from '../../flags/index.js'
@@ -47,9 +48,9 @@ export const info = {
  * @param {string} description
  * @param {readonly string[]} argv
  * @param {ImportMeta} importMeta
- * @returns {void|CommandContext}
+ * @returns {Promise<void|CommandContext>}
  */
-function setupCommand (name, description, argv, importMeta) {
+async function setupCommand (name, description, argv, importMeta) {
   const flags = {
     ...outputFlags,
     ...validationFlags,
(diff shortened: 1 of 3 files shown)


lib/commands/info/index.js
```diff
@@ -90,15 +91,24 @@ function setupCommand (name, description, argv, importMeta) {
...
-  const pkgName = rawPkgName.slice(0, versionSeparator)
-  const pkgVersion = rawPkgName.slice(versionSeparator + 1)
-
-  if (!pkgVersion) {
-    throw new InputError('Need to specify a version, like eg: webtorrent@1.0.0')
+    // Get the latest version
+    try {
+      pkgName = rawPkgName
+      const response = await fetch(`https://registry.npmjs.org/${rawPkgName}/latest`)
+      /** @type any */
```
charliegerard, line 100: Not a fan of using `any` but it seems like there's [an open discussion](https://github.com/node-fetch/node-fetch/issues/1262) on the node-fetch repo about it.

charliegerard:
@mikolalysenko I worked on this before I saw [this related issue in Linear](https://linear.app/socketdev/issue/ENG-978/support-version-aliases-like-latest-in-package-api-resources). I didn't implement it the way it's discussed in there, I would like to work on making the changes in the API to support version aliases but in the meantime, let me know if this PR is still ok or if you'd rather only work on this once the API changes are done.

feross:
@charliegerard thank you for focusing on these developer experience issues! ✨

charliegerard:
Converting back to draft while I work on updating the API instead

bmeck (approved):

PR #98, opened 2024-01-11
Add postinstall script to install the safe npm alias

charliegerard:
## PR description

This PR adds the following functionalities:
 
- A `postinstall` script to ask users if they want to install our npm and npx wrapper so they're automatically protected against malicious packages. (It checks first if the aliases already exist and if so, does not prompt the user)
- The commands `socket wrapper` with the flags `--enable` and `--disable` if users want to add/remove the aliases later on. If the user enters `socket wrapper --enable`, the logic checks if the aliases are already set up and if so, lets the user know, otherwise adds the aliases in the bashrc or zshrc file.
- The commands `socket raw-npm` and `socket raw-npx` to temporarily bypass the wrapper.

Flow:
- When the CLI has finished installing, the output looks like this:

Screenshot 2024-01-10 at 7 39 30 PM

- If the user enters "n", the interface just exits.
- If the user enters "y", the logic checks if the file `~/.bashrc` or `~/.zshrc` exist and if so, the aliases are added to this file and the following output is displayed.

Screenshot 2024-01-10 at 7 44 52 PM


feross:
This is an excellent idea!

Should we also expose it in a CLI command, in case I deny it but want to do it later? Could be a command like `socket wrapper enable` and `socket wrapper disable`?

philliprossii7 (approved):

lib/commands/wrapper/index.js
```diff
@@ -0,0 +1,198 @@
...
+    }
+  })
+}
+
+/**
+ * @param {string} file
+ * @returns {void}
+ */
+const addAlias = (file) => {
+  exec(`echo "alias npm='socket npm'\nalias npx='socket npx'" >> ${file}`, (err, _, stderr) => {
```
bmeck, line 147: we should probably change this to [`fs.appendFile`](https://nodejs.org/api/fs.html#fsappendfilepath-data-options-callback) which avoids a subshell and using `echo` which may not be in the path for some odd minimal containers

lib/commands/wrapper/index.js
```diff
@@ -0,0 +1,198 @@
...
+ * @param {string} query
+ * @returns {void}
+ */
+const askQuestion = (rl, query) => {
+  rl.question(query, (/** @type {string} */ ans) => {
+    if (ans.toLowerCase() === 'y') {
+      try {
+        if (fs.existsSync(BASH_FILE)) {
+          addAlias(BASH_FILE)
+        } else if (fs.existsSync(ZSH_BASH_FILE)) {
```
bmeck, line 127:
```suggestion
        }
        if (fs.existsSync(ZSH_BASH_FILE)) {
```

we should patch both if they exist due to people using different shell wrappers; additionally, we might want to detect the shell?

charliegerard:
Implemented the feedback to replace `echo` with `fs.appendFile` and add the aliases in both the `.bashrc` and `.zshrc` files.

bmeck (approved):

PR #99, opened 2024-01-19
[Fix] Update wrong file path

charliegerard:


Planeshifter (approved):
Fix looks good.

Maybe the variable names with ZSH_BASH_FILE being similar to BASH_FILE, when Zsh and Bash are distinct shells, were to blame for introducing that bug...

bmeck (approved):

PR #103, opened 2024-03-07
disable provenance due npm giving error publishing

bmeck:


.github/workflows/provenance.yml
```diff
@@ -21,6 +21,6 @@ jobs:
          cache: npm
      - run: npm install -g npm@latest
      - run: npm ci
-     - run: npm publish --provenance --access public
+     - run: npm publish
```
Planeshifter, line 24: Think we still need `--access public`:

```suggestion
     - run: npm publish --access public
```

Planeshifter (approved):

PR #114, opened 2024-05-29
Add command to list organizations

charliegerard:
Working towards feature parity with the API, running `socket organizations` will list the organizations associated with the API key used to log in:

Screenshot 2024-05-30 at 4 41 17 PM

If the user is not logged in, running the command will prompt the user for their API key before fetching the organizations.


.github/workflows/nodejs.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       no-lockfile: true
       npm-test-script: 'test-ci'
       # Node 16 has a SegFault from C8 when using --test
-      node-versions: '18,19'
+      node-versions: '19,20'
       # We currently have some issues on Windows that will have to wait to be fixed
```
charliegerard, line 29: The typosquatting test fails in CI with version 18 so not sure why but it passes when updating the versions.

charliegerard:
Converting back to draft as I'm not sure it'd be confusing to have this command when it doesn't actually list all the organizations a user is a member of.

PR #115, opened 2024-05-31
Add cyclonedx command

jdalton:
Add `cyclonedx` command and update deps.


lib/commands/cyclonedx/index.js
```diff
@@ -0,0 +1,208 @@
...
+      },
+      stdout: 'inherit'
+    })`cdxgen ${argvToArray(yargv)}`
+
+    if (cleanupPackageLock) {
+      try {
+        await fs.unlink('./package-lock.json')
+      } catch {}
+    }
+    console.log(chalk.cyanBright(`${yargv.output} created!`))
```
charliegerard, line 206: When running the command `cyclonedx --help`, this message is printed even though the file has not been created.

charliegerard (approved):
Just left a small comment, otherwise looks good!

lib/commands/cyclonedx/index.js
jdalton: Oh that's a good catch. Thank you!

jdalton: Updated to only log if the file exists.

PR #119, opened 2024-06-10
Fix logging module in newer versions of npm

charliegerard:
npm v10.6.0 + doesn't have the module `npmlog` as dependency anymore. Instead, it uses `proc-log` so I updated our wrapper to import the correct module based on the npm version used. Feels like a bit of a hack so let me know if you think of a better way.


bmeck (approved):
seems fine to me

bcomnes (approved):

PR #145, opened 2024-07-18
more robustly handle npm version changes

bmeck:
this fixes modern NPM integration


jdalton (approved):
More or less what my hunch was to fix it. We can add more checks as npm changes things without exact version sniffs.

PR #176, opened 2024-08-20
Small step to prepare for native node glob support, and test waters of dep hygiene

jdalton:
Small step to prepare for native node glob support, and test waters of dep hygiene


jdalton:
@SocketSecurity ignore-all

charliegerard (approved):

PR #311, opened 2025-02-05
Add `socket manifest <lang>` including scala as first lang

pvdz:
This will add a new command to the socket npm cli: `socket manifest <language>`, and the first language: `socket manifest scala`.

Naming is hard, I think `manifest` is most apt, short, and to the point: it creates a dependency manifest.

The command is meant to support things that we currently can't support server side (and also not supported through cyclone (cdxgen)) and which has to run on the client machine.

First candidate is Scala, `sbt` in particular. Since we can't easily find dependencies through static analysis and cyclone doesn't really support it properly, this `socket manifest scala` command would be a wrapper to generate a dependency manifest by levering the (almost guaranteed to exist) local `sbt` setup.

There's also a `socket manifest auto` which tries to detect the proper sub command to execute with default parameters.

- [ ] tests (um)
- [x] do we want to do more with the resulting manifest or leave it like it is now?
  - [ ] we want to figure out the story of how to store these files so we can predictably pick them up through the github api and it must consider multiple languages per repo and mono-repo cases
- [x] somehow verify this is a good approach for actual projects (how?)
- [x] cleanup/formatting/whatevs
- [x] swap out `exec` for `spawn`
- [x] change `sbom` to `manifest`, unless a better name comes up
- [x] think a little harder on the output api, the file to commit to the repo for us to find through github api, what if monorepo, what if polyglot repo, what if multi build targets (dev/prod)
- [x] figure out issue with boolean flags still consuming cli args


src/commands/manifest/index.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  Currently supported language: scala
+
+  Examples
+
+    $ ${name} scala .
+`
+
+export const manifest: CliSubcommand = {
+  description,
```
jdalton, line 31:
```suggestion
  description,
  hidden: true,
```

jdalton: ☝️ This allows us to publish without exposing in the `--help`.

src/utils/meow-with-subcommands.ts
```diff
@@ -9,6 +9,7 @@ import type { Options } from 'meow'
 
 interface CliAlias {
   description: string
+  hidden?: boolean
```
jdalton, line 12: Landed that in `main` ☝️

jdalton (approved):

PR #312, opened 2025-02-07
Remove tap, replace with vitest

pvdz:
Remove tap, replace it with vitest.

Updated tests accordingly.

Inline snapshots here we come!


src/constants.ts
```diff
@@ -126,11 +126,11 @@ const lazyRootDistPath = () =>
   path.join(constants.rootPath, 'dist')
 
 const lazyRootPath = () =>
-  // The '@rollup/plugin-replace' will replace 'process.env.TAP' with `false` and
+  // The '@rollup/plugin-replace' will replace 'process.env.TAP' and VITEST with `false` and
```
pvdz, line 129: um. either we keep TAP "just in case" or update this comment. yes.

jdalton:
Reduces the bug workarounds for tap 🎉 ! Let's see how discussions shake out over the next week!

jdalton (approved):
Okay. Let's do this.

jdalton:
`AbortController` is built-in to Node. It's also a browser standard. Node and its APIs accept the signal
```
const controller = new AbortController();
const signal = controller.signal;
```
as a way to abort requests or operations.

The registry and CLI share an `abortSignal` that is wired through things that will abort when [`'signal-exit'`](https://www.npmjs.com/package/signal-exit) fires its `onExit` callback. This works for things like `Ctrl^C` and other stuff making for clean CLI exits.

We should run Vitest without `JSDOM` if possible.

.config/rollup.dist.config.mjs
```diff
@@ -380,10 +380,10 @@ export default () => {
               )
           ],
           [
-            TAP,
+            VITEST,
             () =>
-              // Lazily access constants.ENV[TAP].
-              !!constants.ENV[TAP]
+              // Lazily access constants.ENV[VITEST].
+              !!constants.ENV[VITEST]
```
jdalton, line 386: No longer need to add this as it's added before.

pvdz: So the TAP one is redundant too?

pvdz: gonna merge. if redundant we can drop it in a separate PR. it doesnt seem to block right now

PR #319, opened 2025-02-11
scan create: allow maven files, include cwd option

pvdz:
This works my way through the `socket scan create` command and ... gets it to work.

Okay. First create a pom.xml and store it somewhere. I've stored it in `~/tmp/okay/pom.xml`

Next, run this:

```
npm exec socket scan create -- --repo=depscantmp --branch=mastertmp --tmp --cwd ~/tmp/okay SocketDev pom.xml
```

(The `npm exec` only for dev stuff)
You'll need to login as usual. The repo and branch dont seem to matter. tmp makes it a temporary scan (?). cwd is mandatory here since in cli dev I can't move to a different folder and various pieces hard require a cwd.

This will lead to an upload like https://socket.dev/dashboard/org/SocketDev/sbom/b48218c6-2d93-40b6-b343-a0dc6e1150df?tab=dependencies

... profit.


src/commands/scan/create.ts
```diff
@@ -141,8 +155,7 @@ async function setupCommand(
     cli.showHelp()
     return
   }
-  const { 0: orgSlug = '' } = cli.input
```
pvdz, line 144: I need to munch on this a bit longer. I think this is backwards breaking. I realize now that this wasn't written by a JS dev and they didn't realize this is not how you destructure an array in JS ...

src/commands/scan/create.ts
```diff
@@ -164,24 +167,46 @@ async function setupCommand(
...
+  const cwd =
+    cli.flags['cwd'] && cli.flags['cwd'] !== 'process.cwd()'
+      ? String(cli.flags['cwd'])
+      : process.cwd()
+
+  const [orgSlug, ...targets] = cli.input
+
   const packagePaths = await getPackageFilesFullScans(
     cwd,
-    cli.input,
```
pvdz, line 169: This would have sent the org as part of potential folders (which probably would lead to an error? or accidental include maybe?). Not sure how blocking this was but it's fixed now.

jdalton (approved):

PR #330, opened 2025-02-18
Refactor many commands into consistency

pvdz:
This refactors a bunch of commands in the same style as https://github.com/SocketDev/socket-cli/pull/322 and some of the others.

I can cut this up in one PR per command if that's easier to process. Ultimately it should not change semantics, only split CLI command handling from the actual thing being executed. And pull the CLI reporting (meow) straight across all commands, where possible.

- [x] is action a github action handler? it's unlisted, do we care/keep/use it?
- [x] how do we safely convert cdxgen to use meow instead of yargs?


src/commands/report/cmd-report-create.ts
```diff
@@ -0,0 +1,120 @@
+import path from 'node:path'
+import process from 'node:process'
+
+import meowOrDie from 'meow'
```
jdalton, line 4: Why is this called `meowOrDie`? and not `meow`?

src/commands/cdxgen/cmd-cdxgen.ts
```diff
@@ -0,0 +1,170 @@
...
+
+import { pluralize } from '@socketsecurity/registry/lib/words'
+
+import { runCycloneDX } from './run-cyclonedx.ts'
+import { getFlagListOutput } from '../../utils/output-formatting.ts'
+
+import type { CliCommandConfig } from '../../utils/meow-with-subcommands.ts'
+
+// TODO: convert yargs to meow. Or convert all the other things to yargs.
+const toLower = (arg: string) => arg.toLowerCase()
```
jdalton, line 14: In this case we're just using `yargsParse`  not full `yargs`. This is essentially a polyfill until Node's [`util.parseArgs`](https://nodejs.org/api/util.html#utilparseargsconfig) is more solid.

src/commands/action/cmd-action.ts
```diff
@@ -0,0 +1,61 @@
+// https://github.com/SocketDev/socket-python-cli/blob/6d4fc56faee68d3a4764f1f80f84710635bdaf05/socketsecurity/socketcli.py
+import meowOrDie from 'meow'
+
+import { runAction } from './run-action.ts'
```
jdalton, line 4: @pvdz The `action` command is the Python-CLI port work. I don't know why it's named `action`.

src/commands/cdxgen/cmd-cdxgen.ts
pvdz: Sure, but we should change that to use meow, right? Be consistent?

src/commands/action/cmd-action.ts
pvdz: At least it's hidden.

My main questionmark is whether it is actually used or not. But us or anyone else. Since it's hidden that's less likely the case. But preferably we just drop this entirely.

It feels to me like it's some kind of GH Actions handler?

jdalton: It's hidden for now but needing more love to be exposed.

src/commands/cdxgen/cmd-cdxgen.ts
jdalton: Yes.

src/commands/report/cmd-report-create.ts
jdalton: Let's call it `meowOrExit`

jdalton (changes requested):
After the naming nit tweaks looks good to land

jdalton (approved):

PR #331, opened 2025-02-18
Fix socket analytics blessed import

pvdz:
I'm not sure who's wrong here but it looks like `import {Widgets} from 'blessed'` doesn't actually work. Or at least it doesn't in our build.

Whichever the way, I've imported it as a type such that we can use it for typing in TS but actual code does a deep import (like it did before).

This fixes it for me.

### DIFF
--- a/src/commands/analytics/display-analytics.ts
+++ b/src/commands/analytics/display-analytics.ts
@@ -1,12 +1,15 @@
 import fs from 'node:fs/promises'
 
-import { Widgets } from 'blessed'
+// @ts-ignore
+import ScreenWidget from 'blessed/lib/widgets/screen'
 import contrib from 'blessed-contrib'
 
 import { Spinner } from '@socketsecurity/registry/lib/spinner'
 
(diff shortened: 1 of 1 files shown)


src/commands/analytics/display-analytics.ts
```diff
@@ -98,7 +101,7 @@ export async function displayAnalytics({
 }
 
 function displayAnalyticsScreen(data: FormattedData): void {
-  const screen = new Widgets.Screen({})
+  const screen: Widgets.Screen = new ScreenWidget({})
```
jdalton, line 104: Would this be implicit by the return type of ScreenWidget? does it need to be explicitly added to `const screen: Widgets.Screen`?

pvdz: TS is ignoring the import so `ScreenWidget` is `any`. But it should be the same value as what I was expecting `Widgets.Screen` to be, except importing it fails because apparently `Widgets` ends up as `undefined`.

This way we get a typed `screen` variable, mitigating the impact of the ts-ignore.

PR #334, opened 2025-02-20
Abstract the meow function

pvdz:
(This depends on https://github.com/SocketDev/socket-cli/pull/333 because Sentry)

Abstract the `meowOrExit` function. This allows us to have a central place to hook into the meows. Case in point: add a tag to Sentry for which command was called to help debugging.

Next PR will add another reasons :d

This PR also cleans up and simplifies the API around calling meow and help etc.


jdalton (approved):
I trust your process

PR #343, opened 2025-02-28
Add `npm run testu` for simpler snapshot updating

pvdz:
[et tu, brute?](https://nl.wikipedia.org/wiki/Et_tu,_Brute%3F)

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,6 @@
     "test:prepare": "cross-env VITEST=1 npm run build",
     "test:unit": "vitest --run",
     "test:unit:coverage": "vitest run --coverage",
-    "test:unit:update-snapshots": "vitest --update",
     "test-ci": "run-s test:*",
     "testu": "npm run build:dist; VITEST=1 npm run build:test-dist ; npm run test:unit -- -u",
     "update": "run-p --aggregate-output update:**",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -63,6 +63,7 @@
     "test:unit:coverage": "vitest run --coverage",
     "test:unit:update-snapshots": "vitest --update",
     "test-ci": "run-s test:*",
+    "testu": "npm run build:dist; VITEST=1 npm run build:test-dist ; npm run test:unit -- -u",
```
jdalton, line 66: Tweak the above line? "test:unit:update-snapshots": "vitest --update"

package.json
```diff
@@ -63,6 +63,7 @@
     "test:unit:coverage": "vitest run --coverage",
     "test:unit:update-snapshots": "vitest --update",
```
jdalton, line 64: ☝️

package.json
pvdz: I don't think we should update snapshots in CI. If something changes we should be aware of it. So I'm not in favor of this one.

pvdz: the above line is `ci`, where I don't think we want to do this... but you made another comment too :p

package.json
pvdz: What is your suggestion here? To add the build step here as well?

It's also not exactly shorter, I may as well be typing out the whole thing at that point ;)

jdalton (approved):

PR #345, opened 2025-02-28
Add suggestion flows to `socket scan create`

pvdz:
This PR updates the `socket scan create` command in a few ways

- Most importantly it adds a suggestion flow
  - When certain flags or args are missing it will use the Socket API to try and make reasonable suggestions
  - When the target is missing it will ask to confirm the target to be the current working directory
  - It will still fail "early" if any of these values are missing after the suggestion flow
- Adds a `--read-only` flag, which is similar to the `--dry-run` flag except it will make REST API calls that only read (eg. no permanent changes)
- Alphabetically orders the function args/params
- Pushes up the error bailout if no api token is present, allowing to better separate the command file logic from the rest
- Fixes `spinner.success` to call `spinner.successAndStop` instead

Here's an example with just running `socket scan create` on the cli folder itself:

image

image

image

image

image

### DIFF
--- a/src/commands/scan/cmd-scan-create.ts
+++ b/src/commands/scan/cmd-scan-create.ts
@@ -2,15 +2,10 @@ import process from 'node:process'
 
 import colors from 'yoctocolors-cjs'
 
-import { Spinner } from '@socketsecurity/registry/lib/spinner'
-
 import { createFullScan } from './create-full-scan'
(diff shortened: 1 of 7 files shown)


jdalton:
I love these enhancements!

jdalton (approved):

PR #348, opened 2025-03-03
Cleanup analytics command, support markdown

pvdz:
This is about the `socket analytics` command:

- Supports the `--markdown` flag, which was previously advertised but completely ignored
  - Including the `--file` flag for it, just like `--json`
- JSON to file will be formatted
- JSON to stdout will be same as the one written to file
  - No more relying on `console.log` to do the formatting, or to hit depth limits
- JSON serialization separately trapped
  - A bit redundant but it's possible for .stringify to hit cicular refs and fail hard
- Fixed up the typing such that the SDK types are properly reflected
- Added some input error conditions which are hidden by default

### DIFF
--- a/src/commands/analytics/cmd-analytics.ts
+++ b/src/commands/analytics/cmd-analytics.ts
@@ -2,13 +2,11 @@ import colors from 'yoctocolors-cjs'
 
 import { logger } from '@socketsecurity/registry/lib/logger'
 
-import { displayAnalytics } from './display-analytics'
+import { displayAnalytics } from './display-analytics.ts'
 import constants from '../../constants'
 import { commonFlags, outputFlags } from '../../flags'
-import { AuthError } from '../../utils/errors'
 import { meowOrExit } from '../../utils/meow-with-subcommands'
 import { getFlagListOutput } from '../../utils/output-formatting'
-import { getDefaultToken } from '../../utils/sdk'
 
 import type { CliCommandConfig } from '../../utils/meow-with-subcommands'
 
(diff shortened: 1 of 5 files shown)


src/commands/analytics/cmd-analytics.ts
```diff
@@ -2,13 +2,11 @@ import colors from 'yoctocolors-cjs'
 
 import { logger } from '@socketsecurity/registry/lib/logger'
 
-import { displayAnalytics } from './display-analytics'
+import { displayAnalytics } from './display-analytics.ts'
```
jdalton, line 5: ☝️ I added a lint rule to main to help auto clean the `.ts` up

src/commands/analytics/display-analytics.ts
```diff
@@ -6,45 +6,32 @@ import contrib from 'blessed-contrib'
 
 import { logger } from '@socketsecurity/registry/lib/logger'
 
+import { fetchOrgAnalyticsData } from './fetch-org-analytics.ts'
+import { fetchRepoAnalyticsData } from './fetch-repo-analytics.ts' // Note: Widgets does not seem to actually work as code :'(
```
jdalton, line 10: ☝️ Widgets comment needs to be moved down

jdalton (approved):

src/commands/analytics/display-analytics.ts
pvdz: yeah that one is annoying. the autofix ignores it.

src/commands/analytics/cmd-analytics.ts
jdalton: It should _(hmm it says it's fixable...)_.

src/commands/analytics/display-analytics.ts
```diff
@@ -305,16 +406,17 @@ function formatData(
...
 }
 
 function renderLineCharts(
   grid: contrib.grid,
-  screen: any,
+  screen: Widgets.Screen,
   title: string,
-  coords: number[],
-  data: { [key: string]: number }
+  coords: Array<number>,
```
jdalton, line 418: ☝️ Added a lint rule to add consistency to types (it's a split the difference approach)

PR #357, opened 2025-03-05
Enable banner by default over stderr

pvdz:
This enables the Socket ascii art banner at the top of every command and help screen.

It looks roughly like this:

```
   _____         _       _        /---------------
  |   __|___ ___| |_ ___| |_      | Socket.dev CLI ver 0.14.57:a70bc02:72eb3827:dev
  |__   | . |  _| '_| -_|  _|     | Node: v20.18.2, API token set: 19tED
  |_____|___|___|_,_|___|_|.dev   | Command: `socket`, cwd: ~/socket/socket-cli
```

This banner contains versioning, login status, command executed (without flags), and the cwd.

This will help us with debugging when people include this in their report.

The banner is emitted over stderr (just like the spinner) and should not interfere with output that's meant to be piped through.

### DIFF
--- a/src/utils/meow-with-subcommands.ts
+++ b/src/utils/meow-with-subcommands.ts
@@ -12,7 +12,7 @@ import { MeowFlags, commonFlags } from '../flags'
 
 import type { Options } from 'meow'
 
-const { DRY_RUN_LABEL, REDACTED, SOCKET_CLI_SHOW_BANNER } = constants
+const { DRY_RUN_LABEL, REDACTED } = constants
 
 interface CliAlias {
   description: string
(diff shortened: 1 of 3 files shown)


src/utils/meow-with-subcommands.ts
```diff
@@ -208,7 +214,7 @@ function getAsciiHeader(command: string) {
     ? REDACTED
     : process
         .cwd()
-        .replace(new RegExp(`^${escapeRegExp(constants.homePath)}`, 'i'), '~/')
+        .replace(new RegExp(`^${escapeRegExp(constants.homePath)}`, 'i'), '~')
```
jdalton, line 217: Ah ya the double `//`. Good catch!

PR #384, opened 2025-03-25
Abstract bad input reporter to make it more consistent

pvdz:
This abstracts the input error reporting to be done by one function.

The advantage is that it looks cleaner on the command files side of things with the output being consistent across all commands. Same coloring, markup, indentation, etc.

Later we'll be able to more easily handle the markdown/json cases, too.


jdalton (approved):

jdalton (dismissed):
Test fails

jdalton (approved):

PR #423, opened 2025-04-07
Fix flags of scan create

pvdz:
Remove the --view flag, which wasn't doing anything
Connect the various commit meta data flags, which were not forwarding the data to the API call

### DIFF
--- a/src/commands/scan/cmd-scan-create.ts
+++ b/src/commands/scan/cmd-scan-create.ts
@@ -59,7 +59,7 @@ const config: CliCommandConfig = {
     dryRun: {
       type: 'boolean',
       description:
-        'run input validation part of command without any concrete side effects'
+        'Run input validation part of command without any concrete side effects'
     },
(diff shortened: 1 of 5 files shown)


src/commands/scan/cmd-scan-create.ts
```diff
@@ -56,34 +56,34 @@ const config: CliCommandConfig = {
       description:
         'Set the default branch of the repository to the branch of this full-scan. Should only need to be done once, for example for the "main" or "master" branch.'
     },
-    pendingHead: {
+    dryRun: {
       type: 'boolean',
-      default: true,
       description:
-        'Designate this full-scan as the latest scan of a given branch. This must be set to have it show up in the dashboard.'
+        'run input validation part of command without any concrete side effects'
```
jdalton, line 62:
```suggestion
        'Run input validation part of command without any concrete side effects'
```

jdalton (approved):

PR #454, opened 2025-04-15
Add `socket manifest conda` to convert to requirements.txt

pvdz:
The `socket manifest conda` will try to extract the pip block from a Conda environment.yml file and return it as a python requirements.txt file.

### DIFF
--- a/src/commands/manifest/cmd-manifest-auto.ts
+++ b/src/commands/manifest/cmd-manifest-auto.ts
@@ -5,6 +5,7 @@ import meow from 'meow'
 
 import { logger } from '@socketsecurity/registry/lib/logger'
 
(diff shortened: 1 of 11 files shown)


mikolalysenko (approved):

src/commands/manifest/cmd-manifest-auto.ts
```diff
@@ -110,6 +111,22 @@ async function run(
     return
   }
 
+  if (existsSync(path.join(dir, 'environment.yml'))) {
```
jdalton, line 114: @pvdz We should also check for `.yaml` too?

pvdz: sure!

PR #458, opened 2025-04-15
Add warning for using `manifest conda`

pvdz:
We don't actually support the Conda ecosystem yet so we're adding a warning for this conversion to try and make sure that it's clear that we don't.


pvdz:
@mikolalysenko please confirm phrasing

BoomchainLabs (approved):

jdalton (approved):

PR #470, opened 2025-04-23
Add node 18 deprecation warning to banner

pvdz:
We should at least give people a heads up about it.

image

(Hm, this would cause the node 18 tests to break because snapshots. Can we maybe just turn that test runner off now in anticipation of the full deprecation?)


jdalton:
Yes

jdalton (approved):

PR #501, opened 2025-05-01
Fix package shallow command due to type mismatch

pvdz:
This fixes `socket package shallow npm babel` which currently breaks due to bad API typing in SDK in OpenAI.


src/commands/package/fetch-purls-shallow-score.mts
```diff
@@ -41,5 +41,9 @@ export async function fetchPurlsShallowScore(
     return handleFailedApiResponse('batchPackageFetch', result)
   }
 
-  return { ok: true, data: result }
+  // TODO: seems like there's a bug in the typing since we absolutely have to return the .data here
+  return {
+    ok: true,
+    data: result.data as SocketSdkReturnType<'batchPackageFetch'>
```
jdalton, line 47: @bcomnes ☝️. @pvdz this happen with the latest sdk update or has it been an issue?

PR #579, opened 2025-05-21
Make manifest auto detect all cases and run any found

pvdz:
This will make `socket manifest auto` first detect all targets and then run any found, rather than only the first one.

This detect script will also be shared with `socket scan create`, later.

### DIFF
--- a/src/commands/manifest/cmd-manifest-auto.mts
+++ b/src/commands/manifest/cmd-manifest-auto.mts
@@ -1,7 +1,5 @@
 import path from 'node:path'
 
-import meow from 'meow'
-
 import { debugLog } from '@socketsecurity/registry/lib/debug'
 import { logger } from '@socketsecurity/registry/lib/logger'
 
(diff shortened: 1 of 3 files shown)


src/commands/manifest/cmd-manifest-auto.mts
```diff
@@ -77,83 +77,59 @@ async function run(
     logger.groupEnd()
   }
 
-  const subArgs = []
-  if (verbose) {
-    subArgs.push('--verbose')
-  }
+  // Both ways work (with and without `=`)
+  // const tmp = meow('', {argv: '--bin x --foo=y'.split(' '), allowUnknownFlags:true, importMeta})
+  // console.log('--->', tmp.input, tmp.flags);
```
jdalton, line 82: ☝️ Debug?

jdalton (approved):
Minor nit.

src/commands/manifest/cmd-manifest-auto.mts
pvdz: mmmm, mental note?

PR #636, opened 2025-06-04
Fix duplicate output in score, add tests

pvdz:
I noticed that the results for a python package were duplicated. This was due to a .release field on the python artifacts, like build targets. I don't think the user cares as much about the score for individual build targets so I'm consolidating those results into a single score, taking the worst value from each result.

The json response will remain raw. Not sure if we want to or should mess with that.

I've also added a test for the score output, shallow and deep, for each ecosystem. At least this way we can track changes when they happen.


jdalton (approved):
🙏 🙏 🐸 🐛

PR #655, opened 2025-06-13
Fix fix command when server rejects request

pvdz:
This requires https://github.com/SocketDev/socket-sdk-js/pull/325

That fix propagates an error response, which is returned as the generator's final (done:true) value. But this value was always ignored.

This PR will pick up the response and even when nothrow is set, when the response is 401 or 404 it will consider it a permanent failure.

(I think it should do this regardless but I'll leave that up to @jdalton to refine)

### DIFF
--- a/src/utils/alerts-map.mts
+++ b/src/utils/alerts-map.mts
@@ -1,8 +1,11 @@
 import { arrayUnique } from '@socketsecurity/registry/lib/arrays'
+import { debugFn } from '@socketsecurity/registry/lib/debug'
+import { logger } from '@socketsecurity/registry/lib/logger'
 
 import { extractPurlsFromPnpmLockfile } from './pnpm.mts'
 import { getPublicToken, setupSdk } from './sdk.mts'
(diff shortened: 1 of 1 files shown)


src/utils/alerts-map.mts
```diff
@@ -12,6 +15,7 @@ import type {
 import type { LockfileObject } from '@pnpm/lockfile.fs'
 import type { Spinner } from '@socketsecurity/registry/lib/spinner'
 
+
```
jdalton:
```suggestion
```

src/utils/alerts-map.mts
```diff
@@ -117,6 +120,19 @@ export async function getAlertsMapFromPurls(
       throw new Error(
         `Socket API server error (${statusCode}): ${statusMessage}`,
       )
+    } else {
+      if (batchResult.status >= 300 && batchResult.status !== 400) {
```
jdalton:
```suggestion
      if (batchResult.status >= 300) {
```

PR #704, opened 2025-08-11
Martin/rea 132 create socket scan create reach option for running tier 1

mtorp:


src/commands/manifest/cmd-manifest-cdxgen.test.mts
```diff
@@ -15,11 +15,8 @@ describe('socket manifest cdxgen', async () => {
         // Need to pass it on as env because --config will break cdxgen
         SOCKET_CLI_CONFIG: '{}',
       })
-      expect(stdout).toMatchInlineSnapshot(
-        `
-        "CycloneDX Generator 11.5.0
-        Runtime: Node.js, Version: 24.5.0"
-      `,
+      expect(stdout).toMatch(
```
pvdz: This change seems unrelated. I'm not expecting vitest to suddenly change this. Can you check and confirm how this change came to be?

src/commands/scan/cmd-scan-create.mts
```diff
@@ -227,6 +234,9 @@ async function run(
     repo: string
     report?: boolean
   }
+  const { reach } = cli.flags as {
+    reach?: boolean
+  }
```
pvdz: There's a `const` destructuring of flags above the `let` destructuring already. Can you please move this one to be part of that one?

src/commands/scan/handle-create-new-scan.mts
```diff
@@ -106,8 +111,29 @@ export async function handleCreateNewScan({
...
+      packagePaths,
+      orgSlug,
+      cwd,
+      repoName,
+      branchName,
+      outputKind,
+      interactive,
+    })
+
+    if (!reachResult.ok || !reachResult.scanPaths) {
```
pvdz: I guess you're doing this because of TS.
If the function returns a CResult<{scanPaths}> then this extra check wouldn't be necessary. I'm expecting this to just do `if (reachResult.ok)`

src/commands/scan/handle-create-new-scan.mts
```diff
@@ -153,3 +179,109 @@ export async function handleCreateNewScan({
...
+  repoName,
+}: {
+  packagePaths: string[]
+  orgSlug: string
+  cwd: string
+  repoName: string
+  branchName: string
+  outputKind: OutputKind
+  interactive: boolean
+}): Promise<{ ok: boolean; scanPaths?: string[] }> {
```
pvdz:
```suggestion
}): Promise<CResult<{ scanPaths: string[] }>> {
```

src/commands/scan/handle-create-new-scan.mts
```diff
@@ -153,3 +179,109 @@ export async function handleCreateNewScan({
...
+  )
+
+  // Lazily access constants.spinner.
+  const { spinner } = constants
+
+  // Setup SDK for uploading manifests
+  const sockSdkCResult = await setupSdk()
+  if (!sockSdkCResult.ok) {
+    await outputCreateNewScan(sockSdkCResult, outputKind, interactive)
+    return { ok: false }
```
pvdz: So how this flow generally ought to work is that you return a CResult here, in this case something like `{ok:false, message: 'reason'}` and then the handler would pass on that result to the output function which will see the ok:false and emit the warning. You're kind of taking a shortcut by calling it directly but should be unnecessary.

pvdz (commented):
Please have a look at how the other commands do the CResult flow; cmd_* -> handle_ -> output_
I'll leave this to @jdalton but leaving some comments to help you along :)

src/commands/manifest/cmd-manifest-cdxgen.test.mts
mtorp: The test was failing if it ran on a system with a Node version different from 24.5.0, but looks like @jdalton made the same observation and fixed the test [here](https://github.com/SocketDev/socket-cli/commit/8aa0304e9f12ed1cc536f330b94dd65928ab8825), so I'm pulling in his changes instead.

src/commands/scan/handle-create-new-scan.mts
mtorp: Thanks Peter!

I believe it's updated to use the idiomatic approach now.

src/commands/scan/handle-create-new-scan.mts
```diff
@@ -106,8 +111,30 @@ export async function handleCreateNewScan({
...
+      repoName,
+      branchName,
+      outputKind,
+      interactive,
+    })
+
+    if (!reachResult.ok) {
+      await outputCreateNewScan(reachResult, outputKind, interactive)
+      return
+    }
```
pvdz, line 131: @mtorp this should just need to return the `reachResult` object.

src/commands/scan/cmd-scan-create.mts
```diff
@@ -93,6 +93,13 @@ const config: CliCommandConfig = {
       description:
         'Similar to --dry-run except it can read from remote, stops before it would create an actual report',
     },
+    reach: {
+      type: 'boolean',
+      default: false,
+      hidden: true,
```
jdalton, line 99: @mtorp Do we want to hide it?

src/commands/scan/cmd-scan-create.mts
```diff
@@ -213,6 +221,7 @@ async function run(
     org: string
     pullRequest: number
     readOnly: boolean
+    reach: boolean
```
jdalton, line 224: ☝️ minor alphabetic order nit... above `readOnly`.

src/commands/scan/cmd-scan-create.mts
mtorp: Just a temporary thing until we get some stuff sorted on the Coana side.

src/utils/coana.mts
```diff
@@ -15,10 +15,20 @@ export async function spawnCoana(
...
+    stdio,
+  } = { __proto__: null, ...options } as SpawnOptions
   const orgSlugCResult = await getDefaultOrgSlug()
   const SOCKET_CLI_API_TOKEN = getDefaultToken()
   const SOCKET_ORG_SLUG = orgSlugCResult.ok ? orgSlugCResult.data : undefined
+
+  // Stop spinner before streaming output if stdio is 'inherit'
+  if (stdio === 'inherit' && spinner) {
+    spinner.stop()
+  }
```
jdalton, line 30: @mtorp ☝️ Nice, good catch!

There's an option to pass the spinner into `spawn` to have it handle that... but it will restore the spinner after I think. No worries. I'll noodle in the code to adjust in either spots as needed.

PR #706, opened 2025-08-12
Martin/rea 137 call the tier1 reachability scanfinalize api from the socket

mtorp:


src/utils/api.mts
```diff
@@ -285,3 +285,101 @@ export async function queryApiSafeJson<T>(
     }
   }
 }
+
+export async function sendApiRequest<T>(
```
mtorp, line 289: This is all Claude.

We can probably make do with something simpler, but made it quite generic in case we need to use it from other places later.

jdalton: That's cool. I may move it to the sdk lib just because it's a generic escape hatch for unpublished apis. Thank you @mtorp !

jdalton:
@mtorp A bonus follow-up would be a test file to test some of the functionality or basics.

jdalton (approved):

PR #708, opened 2025-08-14
Update `socket scan reach` to include the same flags as `socket scan create --reach`

mtorp:
This PR updates `socket scan reach` to:
1. Include the same reachability-specific flags as `socket scan create --reach`.
2. Use the `--manifests-tar-hash` reachability workflow, which is what ensures that SBOMs are computed by Socket instead of the client side SBOM computation from the Coana CLI.

The main difference between `socket scan create --reach` and `socket scan reach` is that `socket scan create --reach` creates a full scan whereas `socket scan reach` just writes the reachability results to `.socket.facts.json`.

### DIFF
--- a/src/commands/scan/cmd-scan-reach.mts
+++ b/src/commands/scan/cmd-scan-reach.mts
@@ -145,12 +145,6 @@ async function run(
   // Accept zero or more paths. Default to cwd() if none given.
   let targets = cli.input || [cwd]
 
-  // Check if we're in dry-run mode first
-  if (dryRun) {
-    logger.log(DRY_RUN_BAILING_NOW)
-    return
-  }
-
   // Use suggestTarget if no targets specified and in interactive mode
   if (!targets.length && !dryRun && interactive) {
     targets = await suggestTarget()
@@ -181,6 +175,11 @@ async function run(
     },
   )
   if (!wasValidInput) {
+    return
+  }
+
+  if (dryRun) {
+    logger.log(DRY_RUN_BAILING_NOW)
     return
   }
 
--- a/src/commands/scan/cmd-scan-reach.test.mts
+++ b/src/commands/scan/cmd-scan-reach.test.mts
@@ -60,7 +60,15 @@ describe('socket scan reach', async () => {
   )
 
   cmdit(
-    ['scan', 'reach', '--dry-run', '--config', '{}'],
+    [
+      'scan',
+      'reach',
+      '--dry-run',
+      '--org',
+      'fakeorg',
+      '--config',
+      '{"apiToken": "abc"}',
+    ],
     'should require args with just dry-run',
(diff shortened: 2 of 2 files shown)


src/commands/scan/perform-reachability-analysis.mts
```diff
@@ -0,0 +1,199 @@
...
+      ? ['--disable-analytics-sharing']
+      : []),
+    ...(reachabilityOptions.reachContinueOnFailingProjects
+      ? ['--ignore-failing-workspaces']
+      : []),
+    // empty reachEcosystems implies scan all ecosystems
+    ...(reachabilityOptions.reachEcosystems.length
+      ? [
+          '--ecosystems',
+          ...convertToCoanaEcosystems(reachabilityOptions.reachEcosystems),
```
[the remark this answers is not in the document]
mtorp: This is wrong!

src/commands/scan/cmd-scan-reach.mts
```diff
@@ -53,30 +95,108 @@ async function run(
...
     logger.log(DRY_RUN_BAILING_NOW)
     return
   }
 
-  const { unknownFlags } = cli
+  // Use suggestTarget if no targets specified and in interactive mode
+  if (!targets.length && !dryRun && interactive) {
+    targets = await suggestTarget()
+  }
+
```
pvdz, line 152: Nit: the dryrun check happens after the `checkCommandInput` call below, usually the last step of these cmd- files.
This way all the tests can test input validation. As long as there is no network activity or write mutations, dry run can continue. Note that `determineOrgSlug` will pass on the dryRun state, for example, and skip network in that case.
Most of the time the cmd- files don't do write mutation or network access (or can do without in dryRun) up to the point where they call the handle file.

src/commands/scan/cmd-scan-create.mts
```diff
@@ -31,42 +32,6 @@ const {
   SOCKET_DEFAULT_REPOSITORY,
 } = constants
 
-const reachabilityFlags: MeowFlags = {
```
pvdz, line 34: Looks odd to me since there's no usage removal but I guess these weren't used before?

pvdz (approved):

src/commands/scan/cmd-scan-reach.mts
mtorp: Thanks, good catch! I've moved the bail now.

src/commands/scan/cmd-scan-create.mts
mtorp: They are in use. Just imported from `reachability-flags.mts` instead so they can be shared between `scan create --reach` and `scan reach`.

jdalton (approved):

PR #709, opened 2025-08-15
Reachability analysis improvements

mtorp:
Contains various bug fixes and improvements
- Fixes a bug where `socket scan create --reach` would use the Docker variant of the Coana analyses.
- Removes the `--reach-continue-on-failing-projects` flag since continuing the scan when the reachability fails on a subproject is now the default behaviour.
- Use the new Coana flag `--purl-types` instead of `--ecosystems` to avoid having to convert Socket purl types to Coana compatible ecosystems.
- Lists the API token requirement for `socket scan reach`.
- Fixes a bug where the bundled reachability analysis for JS would error due to missing dependencies.

### DIFF
--- a/src/commands/scan/cmd-scan-reach.mts
+++ b/src/commands/scan/cmd-scan-reach.mts
@@ -100,7 +100,6 @@ async function run(
     org: orgFlag,
     reachAnalysisMemoryLimit,
     reachAnalysisTimeout,
-    reachContinueOnFailingProjects,
     reachDisableAnalytics,
   } = cli.flags as {
     cwd: string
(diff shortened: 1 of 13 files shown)


src/commands/scan/cmd-scan-reach.mts
```diff
@@ -50,6 +50,10 @@ const config: CliCommandConfig = {
     Usage
       $ ${command} [options] [CWD=.]
 
+    API Token Requirements
+      - Quota: 1 unit
+      - Permissions: full-scans:create
```
jdalton, line 55: ☝️ I love the permission requirements but... I'll try to add that to the error messages too.

jdalton (approved):

PR #742, opened 2025-09-18
Socket fix and repo name adjustments

mtorp:
### Changed
- Rename `--only-compute` flag to `--dont-apply-fixejs` for `socket fix`, but keep old flag as an alias.

### Fixed
- Sanitize extracted git repository names to be compatible with the Socket API.

The Socket backend validates repository names to ensure they work with GitHub's naming rules. However, GitHub's constraints differ from other platforms like GitLab, so names that work on one platform may not work on another. The `sanitizeName` function modifies extracted repository names to be compatible with Socket/GitHub.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [1.1.17](https://github.com/SocketDev/socket-cli/releases/tag/v1.1.16) - 2025-09-18
+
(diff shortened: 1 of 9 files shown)


jdalton:
The 🐉 are now slumbering. All is 🟢

jdalton (approved):

jdalton:
@mtorp Merged to get the PR moving... however `--dont-apply-fixes` we can make it a more conventional `--no-apply-fixes` the negated form of a boolean (it's just a bit of a pain to put into the system). We do it for `--no-banner` and `--no-spinner` though so it's not uncharted.

PR #747, opened 2025-09-25
add --no-major-updates and --show-affected-direct-dependencies flags to socket fix

mtorp:
`--no-major-updates`: Requires that fixes computed by `socket fix` do not change major versions in direct and/or transitive dependencies
`--show-affected-direct-dependencies`: Shows the direct dependencies responsible for introducing transitive vulnerabilities, and show what updates are necessary to remove the vulnerabilities.


jdalton:
Merged via https://github.com/SocketDev/socket-cli/commit/02f3d61a0d8aebea7e685932d7e9e7612c021aea

PR #759, opened 2025-09-30
fix(optimize): correct backwards semver check of node version

Uniqen:
I investigated why no packages was installed when using socket optimize. According to Copilot the issue is that the semver logic in `add-overrides.mts`is backwards. When applying the fix all packages is installed as expected.


jdalton:
Merging in manually.

PR #787, opened 2025-10-11
fix: improve organization capabilities detection for plan variants

dacoburn:
Improves the detection with variants in the capabilities for organizations


jdalton:
pulled in https://github.com/SocketDev/socket-cli/commit/3ac9251e6502c72f811f2a7ed70b009b19d1458f

PR #794, opened 2025-10-16
fix bug where directory target were not handled according to the specification

mtorp:
Running something like `socket scan create  ./path/to/some/folder/with/manifests` previously resulted in the error:
```
✖  Input error:  Please review the input requirements and try again

  ✖ TARGET (file/dir) must contain matching / supported file types for a scan (found no eligible files to scan. See supported manifest files at docs.socket.dev (​https://docs.socket.dev/docs/manifest-file-detection-in-socket​))
```
But the help menu specification for target includes: _If it is a dir, the dir is scanned for any supported manifest files_. I've updated glob.mts such that it now behaves according to the specification.

### DIFF
--- a/scripts/claude.mjs
+++ b/scripts/claude.mjs
@@ -9,8 +9,9 @@ import { existsSync, promises as fs } from 'node:fs'
 import path from 'node:path'
 import { fileURLToPath } from 'node:url'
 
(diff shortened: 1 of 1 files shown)


jdalton:
Merging in manually.

PR #796, opened 2025-10-17
Socket fix improvements

mtorp:
- Adds some e2e tests for `socket fix`. The E2E tests only run in CI since they can be slow and they require a valid `SOCKET_CLI_API_TOKEN` to be set.
- Running `socket fix` with `--json` now outputs the same JSON object as is written to `--output-file <path>`.


jdalton:
Merging in manually.

PR #803, opened 2025-10-20
add `--include` and `--exclude` options for socket fix command

jfblaa:
Related:
https://github.com/coana-tech/coana-package-manager/pull/1911


jdalton:
Merging in manually.

PR #811, opened 2025-10-24
ensure the target directory is handled correctly by the reachability analysis

mtorp:
Fixes a bug where the target directory was ignored by the reachability analysis.
For example, `socket scan reach .` ran the same reachability analysis as `socket scan reach ./foo`.

There are some constraints added to the targets when running in reachability analysis:
- It must be a directory
- It must be inside `cwd`.
- There can only be one target
These constraints are added because the reachability analysis is meant to run on the root of a project.


jdalton:
Manually merged https://github.com/SocketDev/socket-cli/commit/514654f5ad459e5d9647f770e214dddeacf4b32f

PR #850, opened 2025-11-02
Add option --reach-use-unreachable-from-precomputation

barslev:
Adds option `--reach-use-unreachable-from-precomputation` which skips full reachability analysis for the vulnerabilities that precomputed reachability has already classified as unreachable.


jdalton:
Thank you @barslev!

PR #896, opened 2025-11-08
Bump to v1.1.26

jdalton:
Bump to v1.1.26 to include debug logging of API requests/responses

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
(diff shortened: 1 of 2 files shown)


src/commands/analytics/output-analytics.test.mts
```diff
@@ -22,76 +22,76 @@ describe('output-analytics', () => {
             "unmaintained": 133,
           },
           "total_critical_added": {
+            "Apr 18": 0,
```
reberhardt7, line 25: Out of curiosity, why the test change?

jdalton: I've seen the result shift for `analytics` from time to time _(this is a snapshot test so the shift affects it)_. I assume it's some slight variance in the date sorting. It's a [deprecated API](https://docs.socket.dev/reference/getrepoanalytics) so I usually just 🤷  and update the snapshot 🙃

PR #1062, opened 2026-01-28
chore: update telemetry endpoint detection to also match /events

divmain:
This will land alongside larger PRs in depscan and firewall.


jdalton:
Closes as stale, fill free to reopen

PR #1121, opened 2026-03-19
fix: default to cwd when --reach is used without explicit target

mtorp:
## Summary
- Fixed `socket scan create --reach` failing with input validation errors when no explicit target is passed in non-TTY environments (e.g. Jenkins CI). The interactive `select()` prompt silently returns `undefined` when stdin is not a TTY (due to `wrapPrompt` swallowing non-TypeError errors), causing `suggestTarget()` to return `[]` and all reach validations to fail.
- Now defaults to `.` (cwd) when `--reach` is used without a target, consistent with `--reach` requiring exactly one directory target.
- Also fixed latent bug where `cli.input || [cwd]` never fell back to `[cwd]` because empty arrays are truthy in JS.
- Updated @coana-tech/cli to 14.12.200 and bumped CLI version to 1.1.74.

Fixes ENG-4810

## Test plan
- [x] Run `socket scan create --reach --org <org> --repo <repo> --branch <branch>` without explicit target — should succeed using cwd
- [x] Run same command with explicit `.` target — should still work as before
- [x] Run in non-TTY environment (e.g. `echo "" | socket scan create --reach ...`) — should no longer fail with validation errors
- [x] Existing tests pass (24/25 — 1 pre-existing snapshot mismatch unrelated to this change)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Low risk: small CLI input-handling change scoped to `socket scan create --reach` plus a routine dependency/version bump; main impact is defaulting the target to `.` in non-interactive runs.
> 
> **Overview**
> Fixes `socket scan create --reach` when run without an explicit target by correctly detecting an empty `cli.input` array and defaulting the reachability target to `.` instead of falling back to an interactive prompt (which fails in non-TTY CI).
> 
> Bumps the CLI version to `1.1.74`, updates `@coana-tech/cli` to `14.12.200`, and records the change in `CHANGELOG.md`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 572fbcc9fef4617d85f1b14cd0c69961fa44f439. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/commands/scan/cmd-scan-create.mts
+++ b/src/commands/scan/cmd-scan-create.mts
@@ -381,19 +381,20 @@ async function run(
   let updatedInput = false
 
   // Accept zero or more paths. Default to cwd() if none given.
-  // Note: cli.input is always an array (even if empty), so || [cwd] never
-  // fires because [] is truthy. Use .length check instead.
   let targets = cli.input.length ? cli.input : []
 
-  if (!targets.length && reach) {
-    // --reach requires exactly one directory target; default to cwd rather
-    // than relying on an interactive prompt that fails in non-TTY environments
-    // such as Jenkins CI (the select() prompt silently returns undefined when
-    // stdin is not a TTY, causing all downstream validations to fail).
-    targets = ['.']
-  } else if (!targets.length && !dryRun && interactive) {
+  if (!targets.length && !dryRun && interactive) {
     targets = await suggestTarget()
     updatedInput = true
+  }
+
+  // Fallback: if targets is still empty after the interactive prompt (e.g. the
+  // select() prompt silently fails in non-TTY environments like Jenkins CI
+  // because wrapPrompt swallows non-TypeError errors and returns undefined),
+  // default to '.' so that downstream validations don't fail with confusing
+  // "At least one TARGET (missing)" errors.
+  if (!targets.length && !dryRun) {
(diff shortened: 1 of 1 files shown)


src/commands/scan/cmd-scan-create.mts
```diff
@@ -381,9 +381,15 @@ async function run(
...
+  let targets = cli.input.length ? cli.input : []
+
+  if (!targets.length && reach) {
+    // --reach requires exactly one directory target; default to cwd rather
+    // than relying on an interactive prompt that fails in non-TTY environments
+    // such as Jenkins CI (the select() prompt silently returns undefined when
+    // stdin is not a TTY, causing all downstream validations to fail).
+    targets = ['.']
+  } else if (!targets.length && !dryRun && interactive) {
     targets = await suggestTarget()
```
BarrensZeppelin, line 387: Maybe you though about this, but it seems more obvious to me to use '.' as a fallback when the interactive selection fails, instead of always using it when `--reach` is passed.

mtorp: Yes, good point. It's better aligned with how it works without `--reach`.

PR #1197, opened 2026-04-13
fix(hooks): add -- to grep for private key pattern

jdalton:
The private key grep pattern starts with dashes which grep interprets as options. Add -- to signal end of options.

---

> [!NOTE]
> **Medium Risk**
> Changes pre-push secret scanning behavior by inspecting binary files via `strings` and reworking all checks to operate on extracted content, which could introduce new false positives/negatives and unexpectedly block pushes.
> 
> **Overview**
> Improves the `.git-hooks/pre-push` secret scanner by normalizing file inspection: it now detects binary files and scans `strings` output instead of grepping raw bytes, while keeping text files scanned directly.
> 
> Updates all secret/path detectors to operate on the extracted `file_text`, and hardens the private-key check by using `grep --` so the `-----BEGIN ...` pattern isn’t treated as an option.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fa4ebfdac0c85fa4a04557b78d9c09677d82d901. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.git-hooks/pre-push
+++ b/.git-hooks/pre-push
@@ -166,36 +166,51 @@ while read local_ref local_sha remote_ref remote_sha; do
           continue
         fi
 
+        # Use strings for binary files, grep directly for text files.
+        # This correctly extracts printable strings from WASM, .lockb, etc.
+        is_binary=false
+        if grep -qI '' "$file" 2>/dev/null; then
+          is_binary=false
+        else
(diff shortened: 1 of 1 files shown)


jdalton:
Merged into #1196

PR #1221, opened 2026-04-17
chore(install): verify downloaded tarball integrity + enforce HTTPS

jdalton:
## Summary
Small polish pass on `install.sh` so the download path is a little more defensive.

### What changed
- **Integrity check**: after downloading, we compute the SSRI-style hash of the tarball (e.g. `sha512-<base64>`) and compare it to the value the npm registry publishes for the same version. Mismatch aborts before extraction.
- **HTTPS only**: consolidated curl/wget into two helpers that pass `--proto =https --tlsv1.2` (curl) or `--https-only` (wget), so the installer won't follow an http redirect.
- **Temp file via mktemp**: downloads land in a `mktemp` file outside the final install directory, with an `EXIT` trap for cleanup. A failed verify can't leave behind a partial blob that a later run might trust.

### What didn't change
- The user-visible flow, output, and final binary location are identical on a successful install.
- Still supports curl or wget, falls back across `openssl` / `shasum` for the hash.

## Test plan
- [x] Fresh install in a throwaway `$HOME` succeeds and prints "✓ Integrity verified (sha512)"
- [x] Artificially tampering with the downloaded tarball causes the script to abort with a clear mismatch message (verified locally with a one-line patch appending junk to the download)
- [x] `bash -n install.sh` clean
- [ ] CI green

---

> [!NOTE]
> **Medium Risk**
> Changes the installer’s download and verification path (new HTTPS enforcement, temp-file handling, and mandatory integrity checking), which could cause installs to fail in environments missing `openssl`/`mktemp` or with unexpected registry responses.
> 
> **Overview**
> Hardens `install.sh` by routing all network reads/downloads through new `fetch_url`/`fetch_url_to_file` helpers that enforce HTTPS/TLS and prevent redirect-based downgrade.
> 
> Adds an npm-registry integrity verification step: the installer now fetches the published `integrity` for the exact package version, computes the downloaded tarball’s SSRI hash via `openssl`, and aborts before extraction on mismatch or missing metadata.
> 
> Improves safety of the download flow by writing the tarball to a `mktemp`-style temp file outside the install directory and cleaning it up via an `EXIT` trap (plus explicit removal on success).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cec6b1c6b216d47c15acb5e1ac60e544851c3957. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -149,12 +149,27 @@ fetch_url_to_file() {
   fi
 }
 
+# Parse a JSON string field out of a response body. Tolerates a missing
+# field by returning empty, rather than dying under `pipefail`.
+parse_json_string() {
+  local body="$1"
+  local field="$2"
+  # Pipe through `cat` so a grep non-match (exit 1) doesn't trip pipefail;
+  # the final `echo` replaces an empty match with empty string.
+  printf '%s' "$body" \
+    | grep -o "\"${field}\": *\"[^\"]*\"" \
+    | head -1 \
+    | sed "s/\"${field}\": *\"\\([^\"]*\\)\"/\\1/" \
+    || true
+}
+
 # Get the latest version from npm registry.
 get_latest_version() {
   local package_name="$1"
-  local version
+  local body version
 
-  version=$(fetch_url "https://registry.npmjs.org/${package_name}/latest" | grep -o '"version": *"[^"]*"' | head -1 | sed 's/"version": *"\([^"]*\)"/\1/')
+  body=$(fetch_url "https://registry.npmjs.org/${package_name}/latest")
+  version=$(parse_json_string "$body" "version")
 
   if [ -z "$version" ]; then
     error "Failed to fetch latest version from npm registry"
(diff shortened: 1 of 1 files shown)


jdalton:
Both Cursor bugbot findings fixed in 4e3a9e167:

1. **[Low] grep-pipeline exit-on-no-match**: fetch body first, parse via new `parse_json_string` helper that tolerates a missing field under `set -euo pipefail`. The caller's "refusing to install without a published checksum" message now fires with context instead of a silent exit.

2. **[Medium] `xxd` dependency in the shasum fallback**: dropped the fallback entirely and made `openssl` the single required tool. It's ubiquitous on every real target (macOS, mainstream Linux, default Alpine image, WSL, Git Bash); dropping the fallback removes the portability bug and simplifies the function. If openssl is genuinely missing the script prints a platform-specific install hint.

Verified locally:
- happy path still succeeds
- simulated missing-integrity response prints the helpful message
- tampered-download still rejected with expected vs got mismatch

jdalton:
@cursor review

jdalton:
Good catch from @cursor — wget's `--https-only` only applies to recursive downloads, so for single-file fetches it was effectively a no-op. Swapped both wget calls to `--max-redirect=0` in cec6b1c6b, which hard-blocks any HTTPS→HTTP redirect attempt. Verified npm's registry serves both metadata and tarball endpoints with direct 200s (no redirect), so the flag doesn't break the happy path.

Bonus win: this closes a gap in the metadata fetches (version + integrity) that the tarball's own integrity check couldn't cover — a MITM could previously have fed us a fake version + matching fake integrity. Now that vector is blocked at the network layer too.

jdalton:
@cursor review

billxinli (approved):

PR #1227, opened 2026-04-17
fix(fix): validate target directory and detect misplaced IDs

jdalton:
## Summary
Backport of v1.x #1199 to main. Two quality-of-life fixes for \`socket fix\` that surface clear error messages instead of the confusing API error users hit today.

### Misplaced vulnerability ID

Before:
\`\`\`
$ socket fix GHSA-xxxx-xxxx-xxxx
✖ Socket API error: Need at least one file to be uploaded
\`\`\`

After:
\`\`\`
$ socket fix GHSA-xxxx-xxxx-xxxx
✖ "GHSA-xxxx-xxxx-xxxx" looks like a vulnerability identifier, not a directory path.
Did you mean: socket fix --id GHSA-xxxx-xxxx-xxxx
\`\`\`

Triggered on first positional arg matching \`/^GHSA-/\`, \`/^CVE-/\`, or \`pkg:…\`.

### Nonexistent target directory

Before: silently resolve to absolute, glob zero files, API rejects with the same confusing "Need at least one file to be uploaded".

After:
\`\`\`
$ socket fix ./does-not-exist
✖ Target directory does not exist: /absolute/path/to/does-not-exist
\`\`\`

## Tests
- Reworked the existing "should support custom cwd argument" test to use a guaranteed-existing path (\`process.cwd()\`) now that we validate the dir upfront.
- Added "should fail fast when target directory does not exist".
- Added "should fail fast when a GHSA is passed as a positional arg".

## Test plan
- [x] \`pnpm run type\` clean
- [x] \`pnpm --filter @socketsecurity/cli run test:unit\` — all 343 files pass
- [x] \`pnpm run build:cli\` succeeds
- [ ] CI green

---

> [!NOTE]
> **Low Risk**
> Low risk: adds early CLI input validation and clearer error messages, with corresponding unit tests; main risk is minor behavior change for users previously relying on implicit path handling.
> 
> **Overview**
> Improves `socket fix` UX by **detecting GHSA/CVE/PURL values mistakenly passed as the positional CWD** and exiting with a helpful `--id` suggestion instead of proceeding to a confusing upload error.
> 
> Adds **fast-fail validation that the target directory exists** (before org-slug/API token resolution) to surface a clear local error and avoid the API’s generic “Need at least one file to be uploaded” failure. Unit tests are updated/expanded to cover the new validation and ordering guarantees.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e7a1bb1d7b931a4bb31e56dcd17254395fb68b32. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/src/commands/fix/cmd-fix.mts
+++ b/packages/cli/src/commands/fix/cmd-fix.mts
@@ -422,10 +422,18 @@ async function run(
     const isCve = upperInput.startsWith('CVE-')
     const isPurl = rawInput.startsWith('pkg:')
     if (isGhsa || isCve || isPurl) {
-      // `handle-fix.mts` matches GHSA/CVE prefixes case-sensitively, so
-      // echo GHSA/CVE in uppercase in the suggestion. PURLs keep their
-      // original case (they're intentionally lowercase).
-      const suggestion = isPurl ? rawInput : upperInput
+      // `handle-fix.mts` validates IDs with case-sensitive format regexes:
+      //   * GHSA — prefix must be uppercase, body segments lowercase [a-z0-9]
+      //   * CVE  — prefix must be uppercase, body is all digits (case-free)
+      // PURLs are intentionally lowercase and validated separately.
+      let suggestion: string
+      if (isGhsa) {
+        suggestion = 'GHSA-' + rawInput.slice(5).toLowerCase()
+      } else if (isCve) {
+        suggestion = 'CVE-' + rawInput.slice(4)
+      } else {
+        suggestion = rawInput
+      }
       logger.fail(
         `"${rawInput}" looks like a vulnerability identifier, not a directory path.\nDid you mean: socket fix ${FLAG_ID} ${suggestion}`,
       )
--- a/packages/cli/src/commands/fix/cmd-fix.mts
+++ b/packages/cli/src/commands/fix/cmd-fix.mts
@@ -409,6 +409,46 @@ async function run(
     return
   }
 
+  // Detect the common mistake of passing a vulnerability ID (GHSA / CVE /
+  // PURL) as a positional argument when the user meant to use `--id`.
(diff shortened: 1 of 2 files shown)


jdalton:
Both Cursor bugbot findings addressed in e6a91f796:

**[Medium] Misplaced-ID check runs after org-slug network call** — moved both validation blocks (misplaced ID + missing target dir) to fire *before* `getDefaultOrgSlug()`. They depend only on `cli.input[0]`, so users without a configured API token now see the helpful message instead of "Unable to resolve org".

**[Low] Case-insensitive detection suggests command that fails downstream** — `handle-fix.mts` matches GHSA/CVE prefixes case-sensitively, so typing `ghsa-abc` would have produced a broken `socket fix --id ghsa-abc` suggestion. Now we uppercase GHSA/CVE in the suggestion while preserving PURLs (they're intentionally lowercase).

Added three tests covering the uppercase normalization and the two "runs before org-slug resolution" paths.

jdalton:
@cursor review

jdalton:
Good catch from @cursor — the prior fix uppercased the *whole* GHSA ID, but `GHSA_FORMAT_REGEXP = /^GHSA-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}$/` in handle-fix.mts requires the body segments to stay lowercase. So `GHSA-ABCD-EFGH-IJKL` would fail downstream with "Invalid GHSA format".

Fixed in 39978971: only uppercase the prefix.
- GHSA: `'GHSA-' + body.toLowerCase()`
- CVE: `'CVE-' + rest` (body is digits + dashes, case-free)
- PURL: verbatim

Added tests for all three branches plus the PURL passthrough.

jdalton:
@cursor review

jmsjtu (approved):

jdalton:
@cursor review

PR #1229, opened 2026-04-17
fix(pnpm): preserve underscores in package names when stripping peer suffix

jdalton:
## The bug

`stripPnpmPeerSuffix` in `packages/cli/src/utils/pnpm/lockfile.mts` scanned the whole dep path for the first `(` or `_`, which truncated `http_ece@1.0.0` to `http`. That name resolves to npm's `http@0.0.1-security` malware placeholder, so `socket pnpm install` produced a blocking false-positive malware alert on any project that (transitively) depends on `http_ece` — e.g. `web-push@3.6.7`.

### Reproduction

```sh
mkdir /tmp/repro && cd /tmp/repro
echo '{"name":"t","private":true,"dependencies":{"web-push":"3.6.7"}}' > package.json
pnpm install
socket pnpm install --frozen-lockfile
# Before: exits with `http@0.0.1-security: Known malware (critical; blocked)`
# After:  no false positive
```

## Fix

`(` and `_` peer separators only appear *after* the version-separating `@` — semver forbids both characters inside a version, prerelease, or build metadata. Narrow the search to the version-and-beyond portion of the dep path so package-name underscores are preserved.

Walk-through:
- `http_ece@1.0.0_web-push@3.6.7` → scopeOffset=0, versionAt=8, tail=`@1.0.0_web-push@3.6.7`, underscore at tail[6] → slice at 14 → `http_ece@1.0.0` ✓
- `@foo/bar_baz@1.0.0_peer@2.0.0` → scopeOffset=1 (skip scope leader), versionAt=12, tail=`@1.0.0_peer@2.0.0`, underscore at tail[6] → slice at 18 → `@foo/bar_baz@1.0.0` ✓

## Tests

Added two new `it` blocks with six assertions:
- Regression coverage for `http_ece`: `http_ece@1.0.0`, `http_ece@1.2.0_web-push@3.6.7`, `http_ece@1.2.0(web-push@3.6.7)` all preserved through the version.
- Scoped underscore names: `@foo/bar_baz` equivalents prove the scope leader `@` doesn't confuse the version-separator search.

Also rewrote the "prefers parentheses over underscore" test as "prefers the earlier separator when both appear" — observable outcome for the only real-world shape (`(peer)_legacyPeer`) is unchanged but the new name matches the new algorithm.

## Test plan

- [x] `pnpm run type` clean
- [x] `pnpm --filter @socketsecurity/cli run test:unit` — all 343 files / 5158 tests pass
- [x] `pnpm run build:cli` succeeds (pre-commit hooks green)
- [ ] CI green

---

> [!NOTE]
> **Medium Risk**
> Changes parsing used to derive dependency identifiers from pnpm lockfile entries and PURL versions; incorrect edge handling could alter scanned package sets or versions for some projects.
> 
> **Overview**
> Fixes pnpm lockfile parsing by rewriting `stripPnpmPeerSuffix` to only strip peer-dependency suffixes (`(peer@...)` / `_peer@...`) **after the version separator**, preventing underscores in legitimate package names (e.g. `http_ece`) from being truncated and causing false-positive malware matches.
> 
> Extends unit coverage for underscore-in-name packages, scoped packages with underscores, mixed separator ordering, and bare version fragments passed through `resolvePackageVersion`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 97063272db1b8d3afa85a28609a87d0599bbd961. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cli/src/utils/pnpm/lockfile.mts
+++ b/packages/cli/src/utils/pnpm/lockfile.mts
@@ -98,28 +98,95 @@ export function stripLeadingPnpmDepPathSlash(depPath: string): string {
 }
 
 /**
- * Strip pnpm peer dependency suffix from dep path.
+ * Strip pnpm peer dependency suffix from a dep path or bare version.
  *
- * PNPM appends peer dependency info in format:
- * - `package@1.0.0(peer@2.0.0)` - parentheses for peer deps.
- * - `package@1.0.0_peer@2.0.0` - underscore for peer deps.
+ * pnpm appends peer dependency info in one of two shapes:
+ * - `(peer@2.0.0)` — parentheses (pnpm v7+).
+ * - `_peer@2.0.0`  — underscore (pnpm v6 / older lockfiles).
  *
- * Edge cases to consider:
- * - Package names with `(` or `_` in legitimate names (rare but valid).
- * - Scoped packages: `@scope/package` should work correctly.
- * - Empty or malformed dep paths should return unchanged.
+ * The suffix can appear either attached to a full dep path
+ * (`http_ece@1.2.0_web-push@3.6.7`) or to a bare version fragment
+ * (`4.18.0_peer@1.0.0`, passed from `resolvePackageVersion` in
+ * spec.mts). Either way, the separator always sits *after* the
+ * version.
  *
- * Current implementation: Find first `(` or `_` and strip everything after.
- * Limitation: May incorrectly strip if package name contains these chars.
- * Mitigation: Most package names don't contain `(` or `_`, pnpm format is predictable.
+ * Naive `indexOf('_')` is wrong for dep paths because package names
+ * may legitimately contain `_` (e.g. `http_ece`). Searching strictly
+ * after the first `@` is wrong for bare-version inputs because the
+ * only `@` they carry is the one inside the peer suffix itself
+ * (`peer@1.0.0`), so "after the first `@`" lands past the separator.
+ *
+ * Approach:
+ *   1. Find the earliest `_` / `(` in the whole string.
+ *   2. If the first `@` comes before that separator, we have a
+ *      `name@version...` dep path — the separator is a real peer
+ *      suffix and we cut there. (The `@` came from the version,
+ *      not from a peer suffix embedded earlier.)
+ *   3. If there is no `@` before the separator (bare version), the
+ *      separator itself marks the peer suffix and we cut there too.
+ *   4. If there's no separator at all, return unchanged.
+ *
+ * Semver forbids `(` and `_` in a version / prerelease / build
+ * tag, so any `_` or `(` before an eventual version-`@` must be
+ * inside the package name — never a peer separator. That's the
+ * only case where we have to ignore the earliest separator, and it
+ * only applies when the input starts with a name.
+ *
+ * This previously scanned the whole string for `_` / `(` unguarded,
+ * which truncated `http_ece@1.0.0` to `http` and produced a
+ * blocking false-positive alert against npm's `http@0.0.1-security`
+ * malware placeholder.
  */
 export function stripPnpmPeerSuffix(depPath: string): string {
-  // Defensive: Handle empty or invalid inputs.
+  // Defensive: handle empty or invalid inputs.
   if (!depPath || typeof depPath !== 'string') {
     return depPath
   }
 
-  const parenIndex = depPath.indexOf('(')
-  const index = parenIndex === -1 ? depPath.indexOf('_') : parenIndex
-  return index === -1 ? depPath : depPath.slice(0, index)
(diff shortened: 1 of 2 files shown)


jdalton:
Good catch from @cursor — fixed in the latest push (97063272d).

The function has two callers with different input shapes:
- `extractPurlsFromPnpmLockfile` passes dep paths (`name@version...`)
- `resolvePackageVersion` in spec.mts passes bare version fragments (`4.18.0_peer@1.0.0`)

Disambiguator: npm package names can't start with a digit; semver versions always do. Branch on the first character — leading digit means bare version, and the earliest `(`/`_` is the peer-suffix cut. Anything else is `name@version...` and we scan only the version-and-beyond tail (the original fix).

Added a third `it` block covering the bare-version call site: `4.18.0_peer@1.0.0` → `4.18.0`, `4.18.0(peer@1.0.0)` → `4.18.0`, `4.17.21` unchanged. All 343 test files / 5159 tests green.

jdalton:
@cursor review

jdalton:
For context on where `stripPnpmPeerSuffix` is used:

**Two production callers:**

1. **`extractPurlsFromPnpmLockfile`** (`packages/cli/src/utils/pnpm/lockfile.mts:26`) — walks every entry in a parsed pnpm-lock.yaml and converts each dep path to a PURL. This is the path that was producing the false-positive malware alert: a dep path like `http_ece@1.0.0` got truncated to `http` before PURL conversion, and `pkg:npm/http` resolves to the `http@0.0.1-security` placeholder.

   Downstream caller: `getAlertsForLockfile` in `packages/cli/src/utils/socket/alerts.mts:51` — feeds PURLs into the Socket API to fetch alerts for the whole dependency tree. This is the hot path for `socket pnpm install`, `socket scan create`, and anything else that reads pnpm-lock.yaml.

2. **`resolvePackageVersion`** (`packages/cli/src/utils/ecosystem/spec.mts:17`) — takes a `PackageURL` object and returns a clean semver string. Called `stripPnpmPeerSuffix` on the `version` field when `type === 'npm'` to handle pnpm-style `version_peer@x` shapes that leak through into PURL values.

   This is the call site cursor flagged: it passes **bare version strings** (no `name@` prefix), and my first-attempt algorithm broke it by searching for an `@` that wasn't there. The revised algorithm branches on leading-digit to detect bare versions. Note: on current main this function is exported but has no production callers — only unit tests exercise it. Still worth keeping the behavior correct for when a caller is added (or for any external consumers using the package build).

**Test coverage** (35 tests in the suite now):
- Dep-path shapes: plain, parenthesized peer, underscore peer, scoped packages, scoped-with-underscore names, `http_ece` regression cases.
- Bare-version shapes: plain version, version with underscore peer, version with parenthesized peer.
- Edge cases: empty string, null, undefined.

jdalton:
Correction to my earlier context comment — **`socket pnpm` on main uses `sfw`**, not this code path. I missed that.

Verified on main:
- `socket pnpm` → `spawnSfwDlx` (`packages/cli/src/commands/pnpm/cmd-pnpm.mts:92`)
- `pnpm-cli.mts` → `spawnSfw`
- `getAlertsForLockfile` (`utils/socket/alerts.mts:51`) is exported but has zero production callers
- `extractPurlsFromPnpmLockfile` is only called by `getAlertsForLockfile`
- `resolvePackageVersion` (spec.mts) has zero production callers (test-only)

So on current main, `stripPnpmPeerSuffix` has no live call site. This PR's value is narrower than I stated: it fixes a latent bug in a public utility that's currently only exercised by tests, rather than a customer-facing regression. The correctness argument still stands — the function is exported API and the `http_ece` / bare-version shapes should parse correctly — but framing it as a fix for `socket pnpm` was wrong on this branch. It would have been accurate on v1.x where the lockfile path is still live.

jdalton:
Closing: the bug is real but the code path is dead on main.

`socket pnpm` on main routes through `sfw` (`spawnSfwDlx` / `spawnSfw`), not through `utils/pnpm/lockfile.mts`. Tracing the chain:
- `stripPnpmPeerSuffix` — only imported by `extractPurlsFromPnpmLockfile` (same file) and `resolvePackageVersion` (`utils/ecosystem/spec.mts`)
- `extractPurlsFromPnpmLockfile` — only imported by `getAlertsForLockfile` (`utils/socket/alerts.mts`)
- `getAlertsForLockfile` / `resolvePackageVersion` — zero production callers

The customer-visible issue (Rio Transfer, `http_ece` false-positive malware alert) is on v1.x, where the lockfile path is still live. On v3, this utility chain has been replaced by sfw and no longer participates in `socket pnpm install`.

Not worth shipping a fix to code no code path exercises. If the helpers need to come back in main for some other flow, the correctness analysis from the closed PR is still valid.

PR #1232, opened 2026-04-18
fix(cli): return org slug, not display name, from org resolution

jdalton:
## The bug

Orgs whose display name differs from the slug (most obviously, names with spaces like "Example Org Ltd" vs slug `example-org-ltd`) produced broken API URLs when the CLI picked the org automatically or prompted the user to pick one:

```
GET /v0/orgs/Example%20Org%20Ltd/full-scans   → 404
GET /v0/orgs/example-org-ltd/full-scans       → 200
```

Two call sites were returning `.name` when they should have been returning `.slug`.

## Fix

**`commands/scan/suggest-org-slug.mts`** — interactive prompt. Visible label and description still use the friendlier display name; the choice `value` now carries the slug. User sees "Example Org Ltd", we route with "example-org-ltd".

**`commands/ci/fetch-default-org-slug.mts`** — CI auto-resolve. No longer reads `.name` from the first org; uses `.slug`. Also dropped the `Object.keys(organizations)[0]` indirection since `fetchOrganization` already returns an array via `Object.values`.

## Tests

- Added regression guards in both test files exercising the name/slug-differ pattern with a dummy "Example Org Ltd" / `example-org-ltd` fixture.
- Fixed the pre-existing `fetches from API` test in `fetch-default-org-slug.test.mts` — it mocked `organizations` as a dict but the real shape is an array, and asserted the buggy display-name return value. Now asserts slug.
- Renamed `returns error when organization has no name` → `returns error when organization has no slug` with a matching mock payload. The new defensive check guards against a missing slug, not a missing name.

## Test plan

- [x] `pnpm run type` clean
- [x] `pnpm --filter @socketsecurity/cli run test:unit` — 339 files / 5106 tests pass
- [x] `pnpm run build:cli` succeeds
- [ ] CI green

### DIFF
--- a/packages/cli/src/commands/ci/fetch-default-org-slug.mts
+++ b/packages/cli/src/commands/ci/fetch-default-org-slug.mts
@@ -29,8 +29,7 @@ export async function getDefaultOrgSlug(): Promise<CResult<string>> {
   }
 
   const { organizations } = orgsCResult.data
-  const keys = Object.keys(organizations)
-  if (!keys.length) {
+  if (!organizations.length) {
     return {
       ok: false,
       message: 'Failed to establish identity',
@@ -38,10 +37,9 @@ export async function getDefaultOrgSlug(): Promise<CResult<string>> {
     }
   }
 
-  const [firstKey] = keys
-  const slug = firstKey
-    ? ((organizations as any)[firstKey]?.name ?? undefined)
-    : undefined
+  // Use `.slug` (URL-safe) — `.name` is the display label and may
+  // contain spaces ("Example Org Ltd") that break API URLs.
+  const slug = organizations[0]?.slug
   if (!slug) {
     return {
       ok: false,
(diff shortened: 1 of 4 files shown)


jdalton:
@cursor review

jdalton:
@cursor review

jdalton:
@cursor review

jdalton:
@cursor review

jdalton:
@cursor review

jdalton:
@cursor review

packages/cli/src/commands/ci/fetch-default-org-slug.mts
```diff
@@ -29,19 +29,19 @@ export async function getDefaultOrgSlug(): Promise<CResult<string>> {
...
   }
 
-  const [firstKey] = keys
-  const slug = firstKey
-    ? ((organizations as any)[firstKey]?.name ?? undefined)
-    : undefined
+  // Use the API slug (URL-safe identifier), not the display name. Previously
+  // read `.name` which broke API calls for orgs like "Alamos GmbH" whose
+  // display name contains spaces, producing URLs like
+  // `/v0/orgs/Alamos%20GmbH/full-scans` that 404'd.
```
billxinli, line 43: Should these be cleaned up in terms of labels used in the source?

navayla777-dotcom (approved):
[redacted]

billxinli (approved):
LGTM, just a question about the inline comment

PR #1266, opened 2026-04-23
fix(cli): stop socket cdxgen from silently shipping empty-components SBOMs

jdalton:
## What this fixes

`socket cdxgen` can silently produce a CycloneDX SBOM with zero components — a structurally valid document that the Socket dashboard ingests cleanly but has no dependency data to render. The Alerts tab then shows "no alerts," which is indistinguishable from a genuinely clean repo.

Here's how it happens:

When you run `socket cdxgen` without passing `--lifecycle`, we default it to `--lifecycle pre-build` and `--no-install-deps`. That's intentional — it's a safety setting so that generating an SBOM doesn't run arbitrary `postinstall` scripts from the user's dependencies.

But for a Node.js project, that mode needs one of two things to find any dependencies:

1. A lockfile (`pnpm-lock.yaml`, `package-lock.json`, or `yarn.lock`), OR
2. An already-installed `node_modules/` folder.

If the user has neither, cdxgen happily writes out a CycloneDX file with zero components. No error. Our CLI then prints `socket-cdx.json created!` and exits cleanly. The user uploads it, and the dashboard has nothing to show. Nothing in the CLI output tells them anything is wrong.

## What I changed

Two new checks in `socket cdxgen`:

### 1. Hard gate (stops the run before it generates a bad SBOM)

In `cmd-manifest-cdxgen.mts`: right where we apply the default lifecycle, I added a check. If ALL of these are true:

- The user didn't pass `--lifecycle` (so we're using the default),
- The project type is Node.js (the cdxgen default),
- The user didn't pass `--filter` or `--only` (those can intentionally produce empty SBOMs),
- AND there's no lockfile OR `node_modules/` anywhere from cwd upward,

…then we exit with code 2 and a clear error message telling the user exactly how to fix it:

```
socket cdxgen found no lockfile (pnpm-lock.yaml / package-lock.json / yarn.lock)
or node_modules/ at or above <cwd>.
  The default --lifecycle pre-build with --no-install-deps needs one of them
  to resolve components; otherwise the SBOM ships with "components": [].
  Fix: install dependencies first (e.g. `npm install`, `pnpm install`,
  `yarn install`), or re-run with `--lifecycle build` to let cdxgen resolve
  during the build.
```

We walk **up** from cwd (not just check cwd) because a lot of users run this in `packages/foo` inside a monorepo, where the lockfile is at the repo root.

### 2. Soft gate (catches anything the hard gate missed)

In `run-cdxgen.mts`: after cdxgen finishes and writes its output file, we read the JSON back, check if `components` is empty, and if so print a warning. This catches edge cases the hard gate doesn't cover:

- User passed `--lifecycle build` but something went wrong and it still produced zero components.
- Somebody used `--filter` or `--only` too aggressively and every component got filtered out.
- A future cdxgen version changes behavior in a way we didn't predict.

The warning tells them what's wrong and suggests the same fix. Better than them finding out from a blank dashboard.

## How I made the code testable

`run-cdxgen.mts` already had a function that ran cdxgen as a subprocess. I pulled out two small helper functions:

- `detectNodejsCdxgenSources(cwd)` — returns `{ hasLockfile, hasNodeModules }` by walking up from cwd
- `isNodejsCdxgenType(argvType)` — returns true if the `--type` flag is a Node.js platform

Both are exported so the test file can import them. `detectNodejsCdxgenSources` takes an optional `cwd` argument so tests can point it at a specific directory without calling `process.chdir()`, which vitest's worker threads don't allow.

## Tests

Two test files in `packages/cli/test/unit/commands/manifest/`:

**`cmd-manifest-cdxgen.test.mts`** (existing file, added a new describe block):

- Fails when there's no lockfile AND no `node_modules/`
- Passes when only a lockfile is present
- Passes when only `node_modules/` is present
- Skips the gate when the user passes `--lifecycle` explicitly
- Skips the gate for non-Node.js project types (python, java, etc.)

I also had to update the existing `beforeEach` and some existing tests to mock the new helpers so they don't actually hit the filesystem — the mock returns "sources present" by default so the pre-existing tests behave the same as before.

**`run-cdxgen.test.mts`** (new file):

- `isNodejsCdxgenType` across the Node.js types we care about (js, javascript, typescript, nodejs, npm, pnpm, ts)
- `isNodejsCdxgenType` rejects non-Node types (python, java, go, rust)
- `isNodejsCdxgenType` on arrays (if any entry is Node, it matches)
- `detectNodejsCdxgenSources` with various combinations of lockfiles/node_modules present/absent (via mocked `findUp`)

## Test plan

- [x] `pnpm --filter @socketsecurity/cli run type` — clean
- [x] `pnpm --filter @socketsecurity/cli run lint` — clean
- [x] `pnpm --filter @socketsecurity/cli run test:unit test/unit/commands/manifest/cmd-manifest-cdxgen.test.mts` — 38/38 pass (5 new tests for the hard gate)
- [x] `pnpm --filter @socketsecurity/cli run test:unit test/unit/commands/manifest/run-cdxgen.test.mts` — 18/18 pass (new test file)
- [x] `pnpm --filter @socketsecurity/cli run test:unit test/unit/commands/manifest/` — 211/211 pass across the whole manifest folder, no regressions
- [ ] Manual verification in an empty dir: `socket cdxgen` exits 2 with the expected message; with a lockfile, runs normally; with `--lifecycle build`, bypasses the gate.

## What I added to CHANGELOG.md

One entry under `[Unreleased] > Fixed`, explaining the behavior change in user-facing terms.


jdalton:
Good catch — this is a real bug. Fixed in 277662e5f.

**Why the race happened:** `runCdxgen` attached an `async .finally()` to `spawnPromise` but returned the original `cdxgenResult`, not the chained promise. The caller awaited the original, then called `process.exit`. When the async finally yielded at its first `await` (the `fs.readFile` inside `warnIfEmptyComponents`), the caller's continuation fired first and killed the process before the warning printed.

**Fix:** rebind `spawnPromise` on the returned result to the chained promise so `await spawnPromise` in the caller naturally awaits the cleanup + warning. `.finally()` preserves the original value and rejection, so existing callers see identical semantics.

TypeScript note: `.finally()` returns plain `Promise<T>` which strips the `process` / `stdin` extras that `SpawnResult` carries. Callers only use the promise shape (not the extras), so I cast back to `SpawnResult` with a comment explaining why the shape loss is safe.

Type-check clean, full manifest unit suite passes (211/211).

jdalton:
@billxinli @jmsjtu review bump — this one's ready. Bugbot finding resolved, bundle fix for `socket cdxgen` silently emitting empty-components SBOMs. Small diff, one of you good to take a look?

packages/cli/src/commands/manifest/cmd-manifest-cdxgen.mts
```diff
@@ -298,6 +303,29 @@ async function run(
     if (yargv.output === undefined) {
       yargv.output = 'socket-cdx.json'
     }
+
+    // Hard gate: in the default pre-build + install-deps=false path, cdxgen
+    // needs either a lockfile or an installed node_modules/ to produce any
+    // Node.js components. Without both, it emits a valid CycloneDX doc with
+    // "components": []. Refuse with an actionable error instead of shipping
+    // an empty SBOM (SMO-590).
```
billxinli: Should we remove the reference ticket here?

packages/cli/test/unit/commands/manifest/cmd-manifest-cdxgen.test.mts
```diff
@@ -37,8 +37,16 @@ vi.mock('@socketsecurity/lib/logger', () => ({
 
 // Mock runCdxgen to prevent actual cdxgen execution.
 const mockRunCdxgen = vi.hoisted(() => vi.fn())
+const mockDetectNodejsCdxgenSources = vi.hoisted(() =>
+  // Default to "sources available" so pre-existing tests don't trip the
+  // hard gate added in SMO-590.
```
billxinli: same here

packages/cli/test/unit/commands/manifest/cmd-manifest-cdxgen.test.mts
```diff
@@ -311,6 +324,98 @@ describe('cmd-manifest-cdxgen', () => {
       })
     })
 
+    describe('empty-components hard gate (SMO-590)', () => {
```
billxinli: Here too

packages/cli/test/unit/commands/manifest/run-cdxgen.test.mts
```diff
@@ -0,0 +1,90 @@
+/**
+ * Unit tests for run-cdxgen helpers.
+ *
+ * Covers the lockfile/node_modules probe and Node.js type detection that
+ * gate the default `socket cdxgen` path against shipping empty-components
+ * SBOMs (SMO-590).
```
billxinli: one last one

billxinli (approved):
LGTM, just one nitpick on comments

packages/cli/test/unit/commands/manifest/run-cdxgen.test.mts
jdalton: Thanks, the hook that landed will help prevent this

PR #1269, opened 2026-04-24
fix: clarify --default-branch semantics across scan and repository

mtorp:
## Summary

`socket scan create --default-branch` is a boolean toggle, but meow silently discards the value when users pass `--default-branch=<name>` — a common mistake because `socket repository create --default-branch` was a *string* flag that accepted a value. When that happens, the scan is created without being marked as the default branch, so it never lands in the Main/PR dashboard tabs.

- **scan create**: reject `--default-branch=<value>` with an `InputError` that points users at the correct form (`--default-branch` on its own, with `--branch=<name>` to name the branch).
- **repository create / repository update**: rename the flag to `--default-branch-name`. `--default-branch` still works as a hidden meow alias and now emits a deprecation warning pointing at the new name, so existing scripts keep functioning.
- Updated help snapshots and bash completion. Added test coverage for the new error path, the deprecation warning, and the silent-path on the renamed flag.

Resolves ENG-4940.

## Test plan

- [x] `pnpm test:unit src/commands/scan/cmd-scan-create.test.mts src/commands/repository/cmd-repository-create.test.mts src/commands/repository/cmd-repository-update.test.mts` — 43/43 pass (5 new cases)
- [x] `pnpm run check:tsc` — clean
- [x] Manual: `socket scan create --default-branch=main ...` now errors with a clear message
- [x] Manual: `socket repository create fakerepo --default-branch=trunk ...` works and emits deprecation warning
- [x] Manual: `socket repository create --help` shows `--default-branch-name` (old flag hidden from help)


jdalton:
Tackled in vNext as https://github.com/SocketDev/socket-cli/pull/1230. We can punt on v1.x 🕺

PR #1298, opened 2026-05-04
feat(scan): add --exclude-paths flag for full Tier 1 exclusion

simonhj:
## Summary

Adds `--exclude-paths` to `socket scan create` and `socket scan reach` so users can exclude paths from SCA/SBOM manifest discovery and, when reachability runs, from Coana analysis.

The semantics of the flag are scan-root-anchored minimatch-style glob matching, not gitignore-style matching. Each entry is expanded to exclude the matched path itself and its subtree, so `--exclude-paths dist` excludes `<cwd>/dist` and `<cwd>/dist/**`, but not `<cwd>/test/dist`. Folks can use `--exclude-paths '**/dist'` when they want to match `dist` directories at any depth.

SCA already supports gitignore-style ignores from `.gitignore` and `socket.yml` `projectIgnorePaths`, so this PR incurs some extra complexity, contained in `exclude-paths.mts`, to accommodate the difference in semantics. The SCA side receives anchored minimatch ignores, while reachability receives Coana `--exclude-dirs` values re-anchored from the Socket scan root to the current Coana analysis target.

From the UX POV, I think most folks would expect a command-line flag like this to behave as “exclude this path from the scan root, including its contents,” with explicit glob syntax available for broader matching. Not strongly held, so if folks disagree, I’m happy to change it to gitignore-style semantics. We would still need to teach coana how to intepret those, 

---

> [!NOTE]
> **Medium Risk**
> Changes scan file discovery and reachability exclusion behavior, which can materially alter what gets uploaded/analyzed and may unintentionally omit manifests or analysis paths if patterns are misused.
> 
> **Overview**
> Adds a new `--exclude-paths` flag to `socket scan create` and `socket scan reach` to exclude scan-root anchored glob paths from both manifest discovery (SCA/SBOM) and, when tier-1 reachability runs, Coana analysis.
> 
> Introduces `exclude-paths.mts` to validate patterns (no negation/absolute/`..`/match-everything) and to translate the exclusions into (1) fast-glob ignore patterns for manifest collection and (2) target-relative `reachExcludePaths` entries for Coana, including handling nested targets.
> 
> Plumbs these excludes through `handle-create-new-scan`/`handle-scan-reach` and updates globbing utilities (`globWithGitIgnore`, `getPackageFilesForScan`) to accept anchored CLI ignores consistently (including when `.gitignore` negations force the streaming path), with expanded unit/CLI tests and shell completion updates.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 869c848b2170cb491e3d539cf463c80546eff484. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mtorp:
> From the UX POV, I think most folks would expect a command-line flag like this to behave as “exclude this path from the scan root, including its contents,” with explicit glob syntax available for broader matching. Not strongly held, so if folks disagree, I’m happy to change it to gitignore-style semantics. 

No, I agree. I suggest we also deprecate `--reach-exclude-paths` as part of this PR (make it hidden) since it too has different semantics and it's kind of redundant when you have `--exclude-paths`.

mtorp (approved):
LGTM ✅

simonhj:
> > From the UX POV, I think most folks would expect a command-line flag like this to behave as “exclude this path from the scan root, including its contents,” with explicit glob syntax available for broader matching. Not strongly held, so if folks disagree, I’m happy to change it to gitignore-style semantics.
> 
> No, I agree. I suggest we also deprecate `--reach-exclude-paths` as part of this PR (make it hidden) since it too has different semantics and it's kind of redundant when you have `--exclude-paths`.

Done!

PR #1318, opened 2026-05-19
feat(manifest): --facts mode emits Socket facts JSON for Gradle projects (REA-442)

jfblaa:
Closes [REA-442](https://linear.app/socketdev/issue/REA-442/socket-manifest-gradle-facts-generate-socket-facts-json-from-gradle).

## Summary

Adds a `--facts` mode to `socket manifest gradle` / `kotlin` / `auto` that emits a `.socket.facts.json` file describing the resolved compile/runtime dependency graph of an entire Gradle build. Output matches the canonical `SocketFacts` schema consumed by depscan's SBOM_Resolve pipeline and is wired through to coana for tier-1 reachability via `socket scan create . --reach`.

## Highlights

- **New init script** `src/commands/manifest/socket-facts.init.gradle` registers a per-subproject `socketFactsCollect` task plus a build-root `socketFacts` aggregator. The aggregator depends on every collector, emits a **single** `.socket.facts.json` at the build root, and dedupes Gradle's variant artifacts (so a project dep doesn't get split across `java-classes-directory` and `jar` entries). Intra-project deps (`project(':lib')`) are dropped — their externals come through each subproject's own collector. Synchronized accumulators on `gradle.ext` keep `--parallel` safe.
- **No artifact downloads.** Reads `artifact.extension` / `artifact.classifier` from already-fetched POM/GMM metadata; never touches `artifact.file` (which would force Gradle to download every artifact — catastrophic on builds like elasticsearch that declare 300 MB `.deb` distribution archives as deps).
- **`tooling: true` flag** on artifacts that only ever appear outside compile/runtime classpaths (annotation processors, linters, code generators, gradle plugin internals, distribution archive downloads). Downstream reachability scanners can skip them while general alerts still apply. See [REA-445](https://linear.app/socketdev/issue/REA-445) (depscan schema) and [REA-446](https://linear.app/socketdev/issue/REA-446) (coana gate) for the consumer-side companion work.
- **Correct prod/dev classification** with "any non-X clears X" semantics: a dep that appears as both `api` *and* `annotationProcessor` ends up neither `dev` nor `tooling` (production usage wins).
- **`--verbose`** now actually streams Gradle output live (was previously a post-hoc stdout dump).
- **Schema mapping**: `groupId → namespace`, `artifactId → name`, `artifact.extension → qualifiers.ext`, `artifact.classifier → qualifiers.classifier`. Strict on `MavenQualifiersSchema`.
- **Scan-create integration**: the previous filter that stripped `.socket.facts.json` from the `--reach` manifest upload is removed (it predates the producer flow and was breaking it). After a successful scan, the file at the path we instructed coana to write to (`reachResult.data.reachabilityReport`) is unlinked — failed scans leave it for debugging.
- **Surface parity** with the existing pom path: `--facts` on `socket manifest gradle`, `socket manifest kotlin`, and the auto-detect path (via `defaults.manifest.gradle.facts` in `socket.json`); `socket manifest setup` prompts for the toggle.

## End-to-end verification

Tested locally with the full `socket manifest setup` → `socket scan create . --reach --auto-manifest` flow:

1. `socket manifest setup` against the single-module-java fixture (interactive, answered "yes" to the new "(--facts) Emit a Socket facts JSON file?" prompt) → wrote `defaults.manifest.gradle.facts: true` to `socket.json`.
2. `socket scan create . --reach --auto-manifest` → auto-manifest detected gradle → ran our init script → produced `.socket.facts.json` → uploaded to `compute-artifacts` → coana fetched artifacts, ran reachability, wrote enriched output back to `.socket.facts.json` → scan succeeded → cleanup unlinked the file.
3. Repeated the command — every invocation regenerates the facts file end-to-end. The setup → auto-manifest → reach flow is the recommended user pattern for repeatable scans.

## Behavior note for users

Because coana writes its enriched output to `.socket.facts.json` (same path as the producer) and we delete that file after a successful scan, **users who invoke `socket manifest gradle --facts` *manually* will need to re-run it before each subsequent `socket scan create . --reach`**. This is a consequence of keeping the producer and coana output on the same filename, which lets depscan's `**/*.socket.facts.json` glob pick up both shapes without further coordination.

The friction goes away under `--auto-manifest`: each scan run regenerates the facts file via the init script automatically, so the deletion-after-success is invisible.

## Version coverage (manual sdkman matrix)

| JDK | Gradle |
|-----|--------|
| 8 | 5.6.4 |
| 11 | 6.9.4 |
| 17 | 7.6.4 |
| 21 | 8.10.2 |
| 21 | 9.2.1 |

Outputs are byte-identical (after JSON normalization) across the matrix on the test fixtures.

## Test fixtures (added)

- `single-module-java/` — `java-library`, `api` / `implementation` / `testImplementation` deps, plus an annotation processor (lombok) exercising the `tooling: true` flag. Versions pinned (guava 31.1-jre, slf4j 1.7.36) to resolve cleanly under all matrix Gradle versions.
- `multi-module-java/` — root + `lib` + `app` with a `project(':lib')` dep, exercises intra-project drop and variant dedup.
- `unresolved-deps/` — paired real + intentionally-unresolvable dep, exercises lenient `unresolvedModuleDependencies` handling and the `[socket-facts] unresolved: ...` warning.
- `android-library/` — AGP 8.7.3 + compileSdk 34. Exercises per-variant classpaths (`debugCompileClasspath`, `releaseRuntimeClasspath`, ...) and the AGP variant-ambiguity try/catch. Test auto-skips when no `ANDROID_HOME` / `ANDROID_SDK_ROOT` is set.
- `kotlin-multiplatform/` — JVM + JS targets. Exercises per-target classpaths.

## Tests

- `cmd-manifest-gradle.test.mts` / `cmd-manifest-kotlin.test.mts` — `--help` snapshot, `--dry-run`, and `--facts --dry-run` cases. Same shape as existing pom-path tests.
- `socket-facts-init-gradle.e2e.test.mts` — 11 tests across the 5 fixtures, asserts schema shape, direct/dev/tooling attribution, edge integrity, variant dedup, AGP and KMP coverage, single-file emission, and intra-project drop. Auto-skips when no `gradle` binary is on PATH.

## CI

`.github/workflows/e2e-tests.yml` now installs Temurin JDK 21 + Gradle 9.2.1 via the official actions so the integration test actually runs in CI. Android sub-test will skip on CI until `android-actions/setup-android@v3` is added — non-Android coverage is exercised regardless.

## Open follow-ups (separate Linear issues under the same project)

- [REA-445](https://linear.app/socketdev/issue/REA-445) — depscan: schema + consumer support for the `tooling` flag.
- [REA-446](https://linear.app/socketdev/issue/REA-446) — coana: skip `tooling: true` artifacts in the Java reachability resolution loop.

And these remain in the REA-442 description as future work in the producer area:

- Switch to principled discovery via Gradle's data model (`SourceSetContainer`, `androidComponents.onVariants`, `kotlin.targets.compilations`) instead of name-pattern matching. Catches everything we test today, so this is cleanup not coverage.
- Promote the sdkman matrix sweep into a CI job for multi-Gradle coverage.
- Wire Android SDK into CI so the AGP test actually runs there.

## Test plan

- [x] `pnpm test:unit src/commands/manifest` green locally
- [x] `pnpm exec vitest run --config vitest.e2e.config.mts src/commands/manifest/socket-facts-init-gradle.e2e.test.mts` green locally with JDK + Gradle on PATH
- [x] Manual end-to-end: `socket manifest setup` → `socket scan create . --reach --auto-manifest` against a fixture, repeated runs succeed
- [ ] CI green


mtorp:
**Finding 1 — KMP fixture vs Gradle 9.2.1 in CI: version compatibility risk**

`test/fixtures/commands/manifest/gradle-facts/kotlin-multiplatform/build.gradle:7` pins `org.jetbrains.kotlin.multiplatform:1.9.25`. The Kotlin Gradle Plugin 1.9.x officially supports Gradle 6.8.3 → 8.6. CI installs Gradle 9.2.1 (`.github/workflows/e2e-tests.yml:76`), so the KMP test will likely fail on CI even though the local matrix sweep passes (the matrix only goes up to 9.2.1 at JDK 21 — same Gradle version, but local may have a different KGP cache).

Either bump the fixture to Kotlin 2.0.21+ (supports Gradle 9.x) or pin a Gradle version for the e2e job that 1.9.25 supports.

mtorp:
**Finding 2 — Redundant filter pipeline in `handle-create-new-scan.mts`**

At `src/commands/scan/handle-create-new-scan.mts:266-275`:

```ts
const filteredPackagePaths = packagePaths.filter(
  p => path.basename(p).toLowerCase() \!== constants.DOT_SOCKET_DOT_FACTS_JSON,
)
const pathsForScan = reach.reachUseOnlyPregeneratedSboms
  ? filterToCdxSpdxAndFactsFiles(filteredPackagePaths, supportedFiles)
  : filteredPackagePaths
scanPaths = [...pathsForScan, ...(reachabilityReport ? [reachabilityReport] : [])]
```

`packagePaths` has facts files stripped, then we call a function whose name advertises "AndFactsFiles" and which explicitly re-adds them (`handle-create-new-scan.mts:56-60`) — but it can't because they've already been removed. The "include `.socket.facts.json`" branch inside `filterToCdxSpdxAndFactsFiles` is dead in the `--reach` path; the new `reachabilityReport` append handles it instead.

Functionally correct, but the data flow is misleading. Either rename the function to `filterToCdxSpdxOnly` for this call site or reorder the strip/filter so the function's contract still applies.

mtorp:
**Finding 3 — `node.children` mutation guarded only by external lock contract**

In `src/commands/manifest/socket-facts.init.gradle` the per-node `children` is a plain `[] as Set` (HashSet). Writes inside `visit` happen under `synchronized (nodes) { producedIds.each { pid -> nodes[pid].children.addAll(childIds) } }`, which is fine for concurrent writes.

The aggregator (`rootProject { tasks.create('socketFacts') { doLast { ... } } }`) reads `node.children` outside the `synchronized` block. That's safe today because `gradle.projectsEvaluated` wires `aggregator.dependsOn(collector)` for every subproject so all writes happen-before the read — but it's a non-obvious invariant.

Either document the "all writes complete before aggregator runs by virtue of `dependsOn`" guarantee inline, or wrap the aggregator's read in `synchronized(nodes)` for symmetry.

mtorp:
**Finding 4 — `output.stdout.replace(...)` used purely for side effects**

`src/commands/manifest/convert-gradle-to-facts.mts:78-84` uses `String.prototype.replace` purely for the callback side effect, discarding the result. Same pattern is used in the existing `convert_gradle_to_maven.mts`, so it's consistent — but `for (const m of output.stdout.matchAll(/^Socket facts file written to: (.*)/gm)) { logger.log('- ', m[1]) }` reads more clearly.

Cosmetic; happy to defer if you'd rather keep parity with `convert_gradle_to_maven.mts`.

mtorp:
**Finding 5 — Redundant `.toLowerCase()` against an already-lowercase constant**

`src/commands/scan/handle-create-new-scan.mts:267` and `:58` lowercase the basename before comparing to `constants.DOT_SOCKET_DOT_FACTS_JSON`, which is `'.socket.facts.json'` — already lowercase. The `.toLowerCase()` is a no-op.

Trivial cleanup, not blocking.

mtorp:
**Finding 6 — `SOCKET_FACTS_FILENAME` duplicated in Groovy and TS**

`ext.SOCKET_FACTS_FILENAME = '.socket.facts.json'` in `src/commands/manifest/socket-facts.init.gradle` vs `DOT_SOCKET_DOT_FACTS_JSON = \`${DOT_SOCKET_DIR}.facts.json\`` in `src/constants.mts`. Two sources of truth — if one changes, the other silently diverges.

The Groovy script can't import a TS constant, so there's no clean fix. At minimum, leave a short comment in each file pointing at the other so the coupling is visible.

mtorp:
**Finding 7 — Function order vs CLAUDE.md `Function ordering` rule**

`CLAUDE.md` says: "Place functions in alphabetical order, with private functions first, then exported functions." `src/commands/manifest/convert-gradle-to-facts.mts` has `convertGradleToFacts` (exported) before `execGradle` (private/helper). This matches the existing `convert_gradle_to_maven.mts` style though, so it's consistency-with-precedent vs literal-CLAUDE.md.

Trivial — flip both or leave both, just flagging the rule.

mtorp:
**Finding 8 — UX: "Reported exports:" with empty body when no deps**

When the Gradle script hits the `components.isEmpty()` early return (no resolvable deps), it prints `[socket-facts] no resolvable dependencies in build, skipping` but does NOT print a `Socket facts file written to:` line. The TS side in `convert-gradle-to-facts.mts` then logs `Reported exports:` followed by nothing, and still claims `Executed gradle successfully` + suggests `socket scan create`.

Minor UX nit — consider suppressing the `Reported exports:` header when the regex matches zero lines, or surfacing the Gradle skip message so users understand why nothing was emitted.

mtorp (approved):
Approving. Comprehensive, well-documented change with thoughtful Gradle integration (no artifact downloads, parallel-safe accumulators, AGP/KMP coverage, lenient unresolved handling). Left 8 inline-style comments — finding #1 (Kotlin 1.9.25 vs Gradle 9.2.1 on CI) is the only one worth verifying before merge; the rest are nits and follow-up suggestions.

jfblaa:
**Re Finding 1 (KMP/Gradle 9 compat):** Fixed in 71a2a3e06. Bumped the fixture's `org.jetbrains.kotlin.multiplatform` from 1.9.25 to 2.1.0 — KGP 2.1.x supports Gradle 9.x. Smoke-tested locally against Gradle 9.2.1 and the e2e test still produces the expected per-target deps. CI should pick it up.

jfblaa:
**Re Finding 2 (redundant filter pipeline):** Fixed in 71a2a3e06. Renamed `filterToCdxSpdxAndFactsFiles` → `filterToCdxSpdxOnly` and dropped the dead `.socket.facts.json` basename check inside the function body. The function now reads exactly as its name advertises, and the data flow at the call site is honest: strip facts → filter to CDX/SPDX → re-add coana's `reachabilityReport`.

jfblaa:
**Re Finding 3 (`node.children` read outside synchronized):** Fixed in 71a2a3e06. The aggregator's `doLast` now snapshots the accumulator state into plain List/Map values inside a `synchronized (nodes)` block, then builds the JSON outside the critical section. Mirrors the writer-side `synchronized` blocks in each `socketFactsCollect` doLast. Inline comment explains that task dependencies (`aggregator.dependsOn(collector)`) already provide a happens-before guarantee, but the explicit synchronization makes the contract local — no implicit reliance on Gradle's task-graph memory-visibility semantics for plain HashMap/HashSet fields.

jfblaa:
**Re Finding 4 (`replace` for side effects):** Fixed in 71a2a3e06. Switched to `Array.from(output.stdout.matchAll(/^Socket facts file written to: (.*)/gm), m => m[1])` and iterated with a regular `for` loop. Reads more directly and lets me also count matches for Finding 8's fix in the same change. `convert_gradle_to_maven.mts` (the pom path) still uses the older idiom — happy to update that too in a follow-up but left it alone here to keep the PR scope tight.

jfblaa:
**Re Finding 5 (redundant `.toLowerCase()`):** Fixed!

jfblaa:
**Re Finding 6 (`SOCKET_FACTS_FILENAME` duplicated):** Fixed in 71a2a3e06. Added cross-reference comments at both sites:

- `src/commands/manifest/socket-facts.init.gradle` (Groovy `ext.SOCKET_FACTS_FILENAME = '.socket.facts.json'`) now points at `src/constants.mts`
- `src/constants.mts` (TS `DOT_SOCKET_DOT_FACTS_JSON`) now points at the Groovy script

Both comments explicitly call out that the two values are intentionally duplicated (Groovy can't import a TS constant) and must be changed together. As you noted there's no clean fix beyond that; this at least makes the coupling visible.

jfblaa:
**Re Finding 7 (function order):** Not changing in this PR. As you flagged, the existing `convert_gradle_to_maven.mts` puts the exported `convertGradleToMaven` before the private `execGradle` helper — the new `convert-gradle-to-facts.mts` mirrors that ordering deliberately so the two sibling files stay parallel. Reordering both to literal CLAUDE.md compliance is a worthwhile cleanup but feels out of scope here; happy to do it as a separate follow-up if you'd prefer.

jfblaa:
**Re Finding 8 (empty "Reported exports:"):** Fixed in 71a2a3e06. Two parts:

1. `convert-gradle-to-facts.mts` now collects matches via `matchAll` and only prints the `Reported exports:` header when `matchAll.length > 0`.
2. If the match set is empty, we look for the Gradle script's `[socket-facts] no resolvable dependencies` skip message in stdout and surface it via `logger.warn` so the user understands why nothing was written, instead of seeing a silently-empty exports section followed by an inappropriate "next step" suggestion.

PR #1319, opened 2026-05-19
fix(api): always use node:https.request in apiFetch to avoid undici b…

bcomnes:
…ody timeout

Node's built-in fetch uses undici which has a default bodyTimeout of 300s. For large streaming ND-JSON responses (e.g. full scan results) this timeout fires before the body has fully transferred, producing a BodyTimeoutError.

[nodejs/undici` v6.21.3 — `lib/dispatcher/client.js`, line ~246](https://github.com/nodejs/undici/blob/v6.21.3/lib/dispatcher/client.js#L246)

```
this[kBodyTimeout] = bodyTimeout != null ? bodyTimeout : 300e3
```

Route all apiFetch calls through the existing _httpsRequestFetch path, which uses node:https.request and has no body timeout — consistent with the SDK and all other HTTP clients used across socket-cli, socket-sdk-js, and coana.

When SSL_CERT_FILE is not set, agent is passed as undefined and https.request uses its default agent. Behaviour is otherwise identical to before.

Update the apiFetch test to assert https.request is always used and that no custom HttpsAgent is created when CA certs are not configured.

---

> [!NOTE]
> **Medium Risk**
> All CLI API traffic now routes through the custom `https.request` implementation, which could subtly change request/redirect behavior compared to `fetch` and affect all outbound calls. Mitigated by reusing the existing `_httpsRequestFetch` path and tightening tests around agent handling.
> 
> **Overview**
> **`apiFetch` no longer falls back to global `fetch`**; it now always uses the existing `_httpsRequestFetch` (`node:https.request`) path so large/streaming ND-JSON responses aren’t subject to undici’s default body timeout.
> 
> When no extra CA certs are configured, the request is made with `agent: undefined` (default agent); when `SSL_CERT_FILE` provides extra CAs, a custom `HttpsAgent` is still created and passed. Tests were updated to assert `https.request` is always used and that the agent behavior matches both scenarios.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ee86ba4bd33265847c4801b3dda3cb6d32dd6e69. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/utils/api.mts
+++ b/src/utils/api.mts
@@ -20,6 +20,7 @@
  */
 
 import { Agent as HttpsAgent, request as httpsRequest } from 'node:https'
+import { ReadableStream } from 'node:stream/web'
 
 import { messageWithCauses } from 'pony-cause'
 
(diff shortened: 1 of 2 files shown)


src/utils/api.mts
```diff
@@ -186,11 +188,7 @@ export async function apiFetch(
   url: string,
   init: ApiFetchInit = {},
 ): Promise<Response> {
-  const agent = getHttpsAgent()
-  if (!agent) {
-    return await fetch(url, init as globalThis.RequestInit)
-  }
-  return await _httpsRequestFetch(url, init, agent, 0)
+  return await _httpsRequestFetch(url, init, getHttpsAgent(), 0)
```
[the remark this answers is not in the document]
bcomnes: I made a fix

src/utils/api.mts
```diff
@@ -149,29 +151,44 @@ function _httpsRequestFetch(
...
-              statusText: res.statusMessage ?? '',
-              headers: responseHeaders,
-            }),
-          )
+        }
+        // Resolve with a streaming body as soon as headers are available,
+        // matching fetch() semantics. Callers that pipe response.body (e.g.
+        // streamDownloadWithFetch) receive a live ReadableStream rather than
+        // a fully-buffered Buffer.
+        const body = new ReadableStream<Uint8Array>({
```
jdalton, line 170: the v1.x supports Node 18 still so we need to ensure we avoid `ReadableStream` if it is Node 22 and up.

jdalton: Node docs say it was added in Node v18.

jdalton: @bcomnes I think we may need to put an ignore lint comment to fix:
> Error:   169:26  error  The 'ReadableStream' is still an experimental feature The configured version range is '22.19.0'  n/no-unsupported-features/node-builtins
>
>/home/runner/work/socket-cli/socket-cli/src/utils/purl-to-ghsa.mts
Warning:   4:1  warning  `./errors.mts` import should occur before import of `./github.mts`  import-x/order
>
>✖ 2 problems (1 error, 1 warning)

bcomnes: I change it to import out of the package and its not complaining any more.

jdalton (approved):

PR #1331, opened 2026-05-24
fix(scan): finalize tier1 reachability scan from `socket scan reach`

barslev:
## Summary

`socket scan reach` (standalone reachability analysis) spawns Coana, which registers a tier1 reachability scan row via the Socket backend. Until now socket-cli never followed that up with a finalize call from this flow — there was no full-scan id to bind to, so the row stayed at an intermediate post-Coana state forever and looked identical to a stuck `scan create --reach`.

With the tier1 reachability finalize endpoint now accepting a null `report_run_id`, `socket scan reach` can drive the row to its DONE terminal state directly. The intermediate post-Coana state becomes an unambiguous "something went wrong" signal again.

## Changes

- `src/commands/scan/finalize-tier1-scan.mts` — broaden the second argument from `string` to `string | null` (mirrors the finalize endpoint's relaxed schema). No-op for existing call sites that pass a real scan id.
- `src/commands/scan/handle-scan-reach.mts` — after `performReachabilityAnalysis` returns successfully and yields a `tier1ReachabilityScanId`, POST `/tier1-reachability-scan/finalize` with `report_run_id: null`. Failure logs a warn and falls through to the normal output (this is metadata-only; never block the user-visible result on it).
- `src/commands/scan/handle-scan-reach.test.mts` — mock the new dependency and add two focused tests: one verifying `finalizeTier1Scan('tier1-id', null)` is invoked when Coana emits a scan id, one verifying it is NOT called when Coana emits `undefined`.

## Test plan

- [x] `pnpm check:tsc` clean
- [x] `pnpm test:unit src/commands/scan/handle-scan-reach.test.mts` — 6/6 (4 existing + 2 new)
- [x] `pnpm test:unit src/commands/scan/` — 177/177

---

> [!NOTE]
> **Medium Risk**
> Replaces core build/lint/CI and packaging entrypoints, which can break releases or developer workflows if any config mismatch slips through. No direct product logic changes are evident, but the surface area touched (Rollup, ESLint, GitHub Actions, bins) is broad.
> 
> **Overview**
> **Reworks the project’s build + distribution pipeline** by adding new Babel/Rollup configs (including dist bundling that copies/rewires `external/` deps) and switching runtime entrypoints to new `bin/*.js` wrappers that execute the built `dist` outputs.
> 
> **Overhauls developer tooling and CI**: replaces legacy ESLint setup with a flat `eslint.config.js`, adds Biome/Oxlint configs, new `.env.*` profiles, Husky hooks (lint-staged + commit PII guard), and new GitHub Actions workflows for lint/typecheck/test/e2e and npm publish with provenance.
> 
> **Repo hygiene/docs updates**: expands `.gitignore`/`.editorconfig`, removes old `cli.js` + `lib/` command implementation, drops Dependabot config, and refreshes `README.md`/`CHANGELOG.md` plus a new Claude skill for bumping `@coana-tech/cli`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 129adaecf2b92647177bdd5bb03c65e27143be44. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mtorp (approved):
Reviewed against the depscan #20552 server-side change. Integration contract verified:

- Sends literal JSON `null` for `report_run_id` — `sendApiRequest` serializes via `JSON.stringify`, which preserves `null` (not `""`, not an omitted field), matching the relaxed finalize schema. The `string | null` type forbids the `undefined`-drops-the-key footgun.
- Finalize is best-effort: `finalizeTier1Scan` returns a `CResult` and never throws, and a non-ok result only logs a warning before falling through to normal output, so the user-visible result is never blocked.
- Correctly skipped when Coana returns no tier1 scan id; tests assert both the `(id, null)` call and the not-called case.

Two non-blocking notes:
1. Deploy order: this finalize is unconditional (no feature/version gate, by design), so depscan #20552 must be live in prod before v1.1.109 publishes — otherwise today's endpoint 400s on `null` and emits a per-run warning (no command failure).
2. Worth adding a test for the failure path (finalize returns non-ok → warn logged, output still produced).

LGTM.

PR #1334, opened 2026-05-27
feat(manifest): --facts mode emits Socket facts JSON for sbt/Scala projects (REA-474)

jfblaa:
## What

Adds `socket manifest scala --facts`: emit a single `.socket.facts.json` dependency-graph SBOM from an sbt build (mirroring `manifest gradle --facts`, REA-442), for use as pregenerated input to `scan create` Tier 1 reachability instead of generating `pom.xml`.

Closes [REA-474](https://linear.app/socketdev/issue/REA-474).

## How

- **Source-only sbt `AutoPlugin`** shipped in `dist/`, dropped into an isolated `-Dsbt.global.base` plugins dir and run via `sbt -Dsbt.global.base=… socketFacts`. No plugin install, no project changes, never touches the user's `~/.sbt`. Compiled by the sbt meta-build, so a single Ivy-based path spans **sbt 0.13 → 1.x** (no reflection).
- **Metadata-only resolution** via Ivy `ResolveOptions.setDownload(false)` — no artifact jars are downloaded (don't consume client bandwidth). Resolution is per-project, so divergent per-module versions are both reported; intra-build project deps are omitted; one component is emitted per resolved module.
- **Config-scoped for speed + signal:** resolves only the real dependency configs (`compile`/`runtime`/`test`/`provided`/`optional`) by default, skipping the Scala toolchain, sbt-native-packager configs (debian/docker/universal/…), `-internal` duplicates and docs/pom/sources. On lichess/lila (~70 modules): ~48s warm, no downloads.

## Options

- `--configs=compile,test,…` — choose which sbt configurations to resolve.
- `--ignore-unresolved` — skip dependencies that fail to resolve instead of failing the run (default: fail).
- Both also read from `socket.json` (`defaults.manifest.sbt.{configs,ignoreUnresolved}`). The `--facts` toggle is exposed in the `manifest setup` wizard, and `scan create --auto-manifest` honors `defaults.manifest.sbt.facts`.

## Notes

- Why Ivy (not `update`/`updateFull`): `updateFull` resolves *every* configuration (can't be scoped) so it's slower on packager-heavy builds and downloads jars; Ivy's `setConfs` resolves only what we want and stays metadata-only. Coursier's resolve-without-fetch isn't usable here — sbt ships coursier shaded and scalac can't compile against the shaded types ([sbt#6651](https://github.com/sbt/sbt/issues/6651)), and it doesn't exist on 0.13. Ivy resolves the same versions coursier does for real deps and respects `dependencyOverrides` (verified on lila).
- sbt 0.13 requires an older JDK (Java 8/11) — its `TrapExit` calls `setSecurityManager`, which JDK 18+ rejects; the CLI surfaces this with a clear hint.

## Validation

sbt 0.13.18 / 1.9.9 / 1.10.0; Scala 2.11 / 2.13 / 3; single + multi-module; version eviction/overrides; crash-on-unresolved + `--ignore-unresolved`; `--configs` capturing custom (IntegrationTest) configs. `pnpm check` (lint + tsc) and the unit suite pass.

---

> [!NOTE]
> **Medium Risk**
> Runs external `sbt` with a global plugin and depends on correct Ivy resolution metadata for SBOM accuracy; scope is manifest generation only, not auth or data stores.
> 
> **Overview**
> Adds **`socket manifest scala --facts`** to emit a build-root **`.socket.facts.json`** SBOM from sbt (metadata-only Ivy resolution, no `pom.xml`), aligned with existing Gradle facts mode for Tier 1 reachability / `scan create --reach`.
> 
> A new **`SocketFactsPlugin`** Scala source is bundled into **`dist/`** and injected via an isolated **`-Dsbt.global.base`** (no changes to the project or `~/.sbt`). **`convert-sbt-to-facts.mts`** stages the plugin, runs **`socketFacts`**, and surfaces failures (including a JDK hint for legacy sbt on JDK 18+). CLI flags **`--configs`** and **`--ignore-unresolved`** map to JVM properties; defaults can live in **`socket.json`**.
> 
> **`scan create --auto-manifest`**, **`manifest setup`**, help/tests, and the changelog are updated so sbt auto-manifest can prefer facts over `makePom` when configured.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 160208a16e68494fa557aff9b1cb19ed13d284e5. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/commands/manifest/generate_auto_manifest.mts
```diff
@@ -36,19 +37,31 @@ export async function generateAutoManifest({
...
           .filter(Boolean) ?? [],
       verbose: Boolean(sockJson.defaults?.manifest?.sbt?.verbose),
-    })
+    }
+    if (sockJson.defaults?.manifest?.sbt?.facts) {
+      logger.log('Detected a Scala sbt build, generating Socket facts...')
+      await convertSbtToFacts(sbtArgs)
+    } else {
+      logger.log('Detected a Scala sbt build, generating pom files with sbt...')
+      await convertSbtToMaven({
```
mtorp, line 65: `sbtArgs` carries `configs` and `ignoreUnresolved`, which aren't in `convertSbtToMaven`'s parameter type — the spread quietly drops them at runtime because TypeScript's excess-property check doesn't fire on spreads.

Works fine, but it's the kind of thing a future tightening of `convertSbtToMaven`'s signature (or a stricter `exactOptionalPropertyTypes`) would surface as an error, and it makes the call site harder to read than it needs to be.

Suggest either splitting into two arg bags or explicitly destructuring the pom-relevant subset:

```ts
const { configs: _c, ignoreUnresolved: _i, ...pomArgs } = sbtArgs
await convertSbtToMaven({ ...pomArgs, out: ... })
```

src/commands/manifest/cmd-manifest-scala.mts
```diff
@@ -171,6 +223,20 @@ async function run(
...
+  // an explicitly-passed flag. (socket.json defaults don't trip this — only a
+  // flag actually present on the command line does.)
+  if (
+    !facts &&
+    (cli.flags['configs'] !== undefined ||
+      cli.flags['ignoreUnresolved'] !== undefined)
+  ) {
+    logger.warn(
+      'The `--configs` and `--ignore-unresolved` options only apply with `--facts`; ignoring them.',
+    )
```
mtorp, line 237: Nice diagnostic for the `--configs` / `--ignore-unresolved` case — but the inverse (`--out` / `--stdout` passed *with* `--facts`) is silently ignored. The plugin always writes `.socket.facts.json` to the build root; the `socket.outputDirectory` / `socket.outputFile` JVM props aren't exposed by the CLI.

A user who runs `socket manifest scala --facts --out custom.json` will get no warning and no `custom.json`. Suggest a parallel warning here so the symmetry holds:

```ts
if (
  facts &&
  (cli.flags['out'] !== undefined || cli.flags['stdout'] !== undefined)
) {
  logger.warn(
    'The `--out` and `--stdout` options do not apply with `--facts`; the facts file is always written to the build root.',
  )
}
```

src/commands/manifest/socket-facts.plugin.scala
```diff
@@ -0,0 +1,380 @@
...
+      val modules = ivyModule.all(ScopeFilter(inAnyProject)).value
+      val buildRoot = (baseDirectory in ThisBuild).value
+
+      // First pass: every project's own coordinate (org:name), so intra-build
+      // deps are omitted even when referenced as explicit library deps.
+      val projectCoords = mutable.HashSet.empty[String]
+      modules.foreach { module =>
+        module.withModule(log) { (_, md, _) =>
+          val mrid = md.getModuleRevisionId
+          projectCoords += mrid.getOrganisation + ":" + mrid.getName
```
mtorp: `projectCoords` keys on `org:name`, then `isEmittable` (line 274) excludes any external dep that matches. If an external dependency on Maven Central happens to share `org:name` with a project in this build (e.g. a project named `org.example:utils` and an unrelated external `org.example:utils`), it'll be silently filtered out with no warning.

Vanishingly small risk in practice, but it's also a silent failure mode — a missing component with no diagnostic. Consider either:

- Keying on `org:name:version` (still imperfect but tighter), or
- Logging a `[socket-facts] skipping intra-build coordinate <coord>` when the filter drops a node, so a surprising hit at least surfaces in the run output.

src/commands/manifest/cmd-manifest-scala.mts
```diff
@@ -28,6 +29,21 @@ const config: CliCommandConfig = {
...
     },
+    facts: {
+      type: 'boolean',
+      description:
+        'Emit a Socket facts JSON file (`.socket.facts.json`) describing the resolved dependency graph instead of generating `pom.xml` files',
+    },
+    configs: {
+      type: 'string',
+      description:
+        'With --facts: comma-separated sbt configurations to resolve (default: compile,runtime,test,provided,optional)',
```
mtorp: Nit: the description here lists the defaults as `compile,runtime,test,provided,optional`, but `DefaultConfs` in `socket-facts.plugin.scala:72-73` lists them alphabetically: `compile, optional, provided, runtime, test`. Same set, different order.

Not a bug, but the two will drift the next time someone touches one and not the other. Worth picking one ordering and using it in both spots.

mtorp (approved):
Nice work. The Ivy-based, source-only plugin approach is well-chosen, the cross-version (sbt 0.13 ↔ 1.x) story holds up, and the comments are doing real load-bearing work. Approving — left four small inline notes (none blocking) on places where the sbt path quietly diverges from the existing Gradle facts mode in user-visible or code-cleanliness ways.

src/commands/manifest/generate_auto_manifest.mts
jfblaa: Fixed in 367127ff0. The shared `sbtArgs` bag now only holds the fields both handlers accept (`bin`, `cwd`, `sbtOpts`, `verbose`). The facts-only `configs`/`ignoreUnresolved` are added in the `--facts` branch and the pom-only `out` in the `else` branch, so `convertSbtToMaven` is no longer spread properties it doesn't declare — and the call sites read more clearly. I went with two arg-bag extensions rather than the destructure-discard so there are no unused `_c`/`_i` bindings to placate the linter.

src/commands/manifest/cmd-manifest-scala.mts
jfblaa: Good call — added the symmetric guard in 367127ff0:

```ts
if (
  facts &&
  (cli.flags['out'] !== undefined || cli.flags['stdout'] !== undefined)
) {
  logger.warn(
    'The `--out` and `--stdout` options do not apply with `--facts`; the facts file is always written to the build root.',
  )
}
```

So `socket manifest scala --facts --out custom.json` (or `--stdout`) now warns instead of silently producing nothing there. It reads raw `cli.flags` (not the defaulted vars) so a `socket.json` `outfile` default doesn't trip it — same convention as the `--configs`/`--ignore-unresolved` guard above.

src/commands/manifest/socket-facts.plugin.scala
jfblaa: Fixed in 367127ff0 by keying on the full GAV. Added a shared helper:

```scala
private def gavKey(mrid: ModuleRevisionId): String =
  mrid.getOrganisation + ":" + mrid.getName + ":" + mrid.getRevision
```

used both when collecting project coordinates (first pass) and in `isEmittable`. An external dependency that merely shares an `org:name` with a build project — at a different version — is now emitted. After this the only way the filter drops a node is an exact `org:name:version` match with an in-build project, which can't resolve to a distinct external artifact, so the silent-drop window the comment flagged is closed.

I chose the tighter key over the per-node "skipping intra-build coordinate" log because the latter would be noisy on large multi-module builds (every sibling reference would log a line), and GAV-keying removes the realistic collision entirely rather than just surfacing it.

src/commands/manifest/cmd-manifest-scala.mts
jfblaa: Fixed in 367127ff0 — sorted the help-text default list to match `DefaultConfs`: `compile,optional,provided,runtime,test` (alphabetical, the repo's list-ordering convention). Inline help snapshot regenerated.

PR #1352, opened 2026-06-02
feat(manifest): default to Socket facts, delegate generation to Coana CLI

jfblaa:
## Summary

Make `socket manifest {gradle,kotlin,scala,auto}` emit `.socket.facts.json` by default and add `--pom` for the legacy pom.xml generation.

- Facts generation now delegates to the Coana CLI's `manifest gradle|sbt` command via `spawnCoanaDlx` (which honors `SOCKET_CLI_COANA_LOCAL_PATH`). socket-cli no longer runs gradle/sbt itself for the facts path; the build-tool resolution scripts live in Coana.
- New shared helper `coana-manifest-facts.mts`; `convert-gradle-to-facts` / `convert-sbt-to-facts` are now thin wrappers.
- **Default flip:** facts is the default; `--pom` opts into pom.xml generation. `--facts` is still accepted (it's the default); socket.json `facts: false` still selects pom; passing `--facts --pom` together warns and generates facts. `socket manifest auto` flipped to match.
- Removed the now-dead bundled `socket-facts.init.gradle`, the sbt plugin, their rollup copy steps, and the `gradle-facts` test fixtures (ported to Coana). The pom-path `init.gradle` is unchanged.

The resolved artifact-paths sidecar (file paths used for reachability) stays internal to Coana — `socket manifest` only requests the uploadable facts file.

## Testing

- `check:tsc`, `check:lint`, Biome clean; full unit suite passing (help snapshots + auto-manifest tests updated).
- Live: `socket manifest gradle` (no flags) against a Gradle (Fabric Loom) test project delegates to a local Coana build and produces the expected `.socket.facts.json`.

Linear: REA-507 (related: REA-509)

---

> [!NOTE]
> **High Risk**
> Default output flipped from pom.xml to facts and facts now depend on Coana `manifest` subcommands that may not ship in the pinned CLI version yet, so this is a breaking UX change with an external runtime dependency.
> 
> **Overview**
> **`socket manifest gradle`, `kotlin`, and `scala`** now emit **`.socket.facts.json` by default**; **`--pom`** opts into the previous **`pom.xml`** path. **`--configs`** is replaced by **`--include-configs`** and **`--exclude-configs`**. **`socket manifest auto`**, **`socket.json` defaults**, and the **setup wizard** follow the same default and flag names.
> 
> Facts generation no longer runs Gradle/sbt or ships in-repo resolution scripts. A new **`coana-manifest-facts.mts`** helper forwards to **`coana manifest gradle|sbt`** via **`spawnCoanaDlx`**; the Gradle/sbt facts converters are thin wrappers. **Bundled `socket-facts.init.gradle` and the sbt facts plugin**, their **rollup copy steps**, and **local gradle-facts fixtures** are removed. The **pom** **`init.gradle`** path is unchanged. Release **1.1.113** bumps **`@coana-tech/cli`** to **15.3.19**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit acdba639e02f96b6711215673605667d46bd9fee. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/commands/manifest/cmd-manifest-gradle.mts
```diff
@@ -32,12 +32,22 @@ const config: CliCommandConfig = {
...
+      description:
+        'When generating facts: comma-separated glob patterns matched against Gradle configuration names (case-sensitive, `*` and `?` wildcards). Only configurations matching at least one pattern are resolved. e.g. `*CompileClasspath,*RuntimeClasspath`. Default: every resolvable configuration except AGP instrumented-test classpaths',
+    },
+    excludeConfigs: {
       type: 'string',
       description:
-        'With --facts: comma-separated glob patterns matched against Gradle configuration names (case-sensitive, `*` and `?` wildcards). e.g. `*CompileClasspath,*RuntimeClasspath` to skip tooling configs. Default: every resolvable configuration except AGP instrumented-test classpaths',
+        'When generating facts: comma-separated glob patterns; Gradle configurations matching any pattern are skipped (applied after --include-configs)',
     },
     ignoreUnresolved: {
```
mtorp, line 52: Nit (consistency): now that facts is the default, this `--ignore-unresolved` description (line 55) still reads _"With --facts: ..."_ while the adjacent `--include-configs` / `--exclude-configs` were reworded to _"When generating facts: ..."_. Worth aligning the wording. The same `--ignore-unresolved` string also appears in `cmd-manifest-kotlin.mts` and `cmd-manifest-scala.mts` (and their help snapshots), so they'd need updating together.

src/commands/manifest/coana-manifest-facts.mts
```diff
@@ -0,0 +1,90 @@
...
+// init script and the sbt plugin) live in Coana now, so socket-cli no longer
+// runs them itself; it only asks Coana for the uploadable `.socket.facts.json`.
+//
+// The resolved artifact-paths sidecar is intentionally NOT requested here: it
+// only matters for reachability analysis, which is internal to Coana, so Coana
+// emits it itself when it runs reachability. `socket manifest` only needs the
+// facts file.
+//
+// `spawnCoanaDlx` resolves the Coana CLI via dlx (or a local build when
+// `SOCKET_CLI_COANA_LOCAL_PATH` is set). `bin` (the gradle/sbt executable) is
```
mtorp, line 20: Nit (comment accuracy / dead branch): the helper forwards `--bin` only via `if (bin)`, and this comment says _"when empty, Coana applies the … default (`./gradlew`, or `sbt` on PATH)"_. But every caller — `cmd-manifest-{gradle,kotlin,scala}.mts` and `generate_auto_manifest.mts` — already defaults `bin` to `<cwd>/gradlew` or `sbt` before calling, so `bin` is never empty here and the "when empty" path is effectively dead. Not a bug (the forwarded defaults match Coana's), but either the comment overstates reality or the empty-guard could be dropped.

src/commands/manifest/coana-manifest-facts.mts
```diff
@@ -0,0 +1,90 @@
...
+    undefined,
+    { cwd },
+    { stdio: 'inherit' },
+  )
+  if (!result.ok) {
+    process.exitCode = 1
+    logger.fail(result.message || 'Coana failed to generate Socket facts')
+    return
+  }
+  logger.success('Generated Socket facts')
```
mtorp, line 119: Nit (messaging): this prints `logger.success('Generated Socket facts')` + the next-step line unconditionally on any `result.ok`. The old converters detected the _"no resolvable dependencies"_ skip and warned instead. Now that detection lives in Coana, a Coana run that exits 0 but writes no facts file would still print a success/next-step message here. Fine if Coana always signals the no-deps case via a non-zero exit or its own output; flagging in case the unconditional success line can mislead.

src/commands/manifest/cmd-manifest-scala.mts
```diff
@@ -225,17 +268,19 @@ async function run(
...
   if (
     !facts &&
-    (cli.flags['configs'] !== undefined ||
+    (cli.flags['includeConfigs'] !== undefined ||
+      cli.flags['excludeConfigs'] !== undefined ||
       cli.flags['ignoreUnresolved'] !== undefined)
   ) {
     logger.warn(
-      'The `--configs` and `--ignore-unresolved` options only apply with `--facts`; ignoring them.',
+      'The `--include-configs`, `--exclude-configs`, and `--ignore-unresolved` options only apply when generating Socket facts (not with `--pom`); ignoring them.',
```
mtorp, line 284: Two notes on the `--out`/`--stdout` block just below (lines 287–299, unchanged by this PR):

1. **Stale comment** — lines 287–291 still reference _"the plugin"_ and _"socket.outputDirectory/outputFile JVM props"_, which moved into Coana; socket-cli no longer owns an sbt plugin, so the rationale reads out of date.
2. **UX (design call, non-blocking)** — with facts now the default, `socket manifest scala --out foo.xml` (no `--pom`) silently generates facts and warns that `--out`/`--stdout` don't apply. A user passing `--out`/`--stdout` almost certainly wanted pom output. Consider treating those flags as implying `--pom`, or keep requiring explicit `--pom` if that's the intended contract.

mtorp (commented):
A few minor, non-blocking nits from the review (docs/comments + one UX design question). Nothing here blocks merge — the two larger items (the default-output flip and the Coana 15.3.19 runtime dependency) are intentional/documented and I'll raise those separately.

mtorp (approved):
Approving the code. Verified locally: `check:tsc`, ESLint, and the manifest unit suite (22/22) all pass; build is clean. Traced the facts/`--pom` default-flip logic and the `generateAutoManifest` opt-out across all branches — correct. All converter call sites updated, no dangling refs to the removed init.gradle/sbt-plugin/fixtures, version pin consistent.

Left a few minor doc/comment nits inline; none are blocking.

src/commands/manifest/cmd-manifest-gradle.mts
jfblaa: Done — aligned all three `--ignore-unresolved` descriptions (gradle/kotlin/scala) to the "When generating facts: …" phrasing used by the adjacent `--include-configs`/`--exclude-configs`, and regenerated the help snapshots.

src/commands/manifest/coana-manifest-facts.mts
jfblaa: Good catch. Every caller (`cmd-manifest-{gradle,kotlin,scala}.mts` and `generate_auto_manifest.mts`) resolves `bin` to a concrete default (`<cwd>/gradlew` or `sbt`) before calling, so the "when empty" path was indeed dead. I reworded the comment to say `bin` is always resolved by the caller and forwarded verbatim, and kept the `if (bin)` guard only as a cheap safeguard against accidentally passing `--bin ""`.

src/commands/manifest/coana-manifest-facts.mts
jfblaa: Intentional — with generation delegated to Coana we now treat Coana's exit code as the source of truth. Coana exits non-zero when it can't produce facts, and otherwise streams its own "Socket facts file written to: …" line, so this success/next-step message tracks Coana's own outcome rather than re-detecting state we no longer compute. The old per-converter "no resolvable dependencies" detection was tied to the in-process resolution that has moved into Coana, so it can't live here anymore. Happy to revisit if we find a case where Coana exits 0 without writing a facts file.

src/commands/manifest/cmd-manifest-scala.mts
jfblaa: Thanks — addressed note 1 and kept note 2 as-is by design:

1. **Stale comment** — fixed. It no longer references "the plugin" or the `socket.outputDirectory/outputFile` JVM props (those moved into Coana). It now just notes that when generating facts, Coana always writes `.socket.facts.json` to the build root and its output location isn't exposed through this CLI.

2. **`--out`/`--stdout` implying `--pom`** — kept the explicit-`--pom` contract for now. Inferring `--pom` from an output-location flag would silently flip the output *format* based on an unrelated flag, which felt more surprising than a no-op; the warning makes the no-op visible instead. Open to revisiting if this trips users up in practice.

src/commands/manifest/coana-manifest-facts.mts
jfblaa: Following up — you were right to flag this; trusting Coana's exit code alone was the wrong call. Coana writes facts to a single deterministic path (`<cwd>/.socket.facts.json`), so I now verify that file exists after a `result.ok` run before printing the success/next-step line. If Coana exits 0 without producing a facts file (the no-resolvable-dependencies case), we now `logger.warn` "Coana completed but wrote no .socket.facts.json (no resolvable dependencies?); nothing to upload." and skip the misleading next-step message, rather than assuming the exit code told the whole story.

src/commands/manifest/cmd-manifest-scala.mts
jfblaa: Follow-up on note 2 — we firmed up the contract instead of leaving it a soft warning. `.socket.facts.json` is a transient artifact that `socket scan create <project>` discovers in the project root, so letting `--out`/`--stdout` redirect it would just decouple it from where the next step looks. So:

- `--out`/`--stdout` now **hard-fail** (usage error, exit 2) when generating Socket facts, rather than silently warning — `The \`--out\` and \`--stdout\` options only apply with \`--pom\`; Socket facts are always written to the project root as \`.socket.facts.json\``. Added a regression test for it.
- Their `--help` descriptions now say "Only with --pom: … Does not apply when generating Socket facts".
- Note 1's stale comment is reworded to "project root" (it's derived from the path argument, e.g. `socket manifest gradle /path/to/project` writes into that project — not CWD).

Reachability is unaffected and stays consistent: the *initial* facts file goes through the same `runCoanaManifestFacts` (project root, no override), while the *final* reachability report is a separate `coana run --socket-mode` output that `socket scan reach --output` can still redirect.

PR #1363, opened 2026-06-15
feat(manifest): discover and scan nested Bazel sub-workspaces

simonhj:
## What

Adds nested **sub-workspace discovery** to `socket manifest bazel`. Today the
command extracts Maven dependencies from a single workspace rooted at the scan
directory; a repo that nests additional Bazel workspaces (e.g. a `mobile/`
sub-tree with its own `MODULE.bazel`) only ever gets its top-level workspace
scanned, so the nested workspace's dependencies are missed.

This change walks the directory tree, finds every directory containing a
workspace marker (`MODULE.bazel` / `WORKSPACE` / `WORKSPACE.bazel`), and runs
the existing single-workspace extractor once per discovered root, writing one
manifest per workspace at a path mirroring its location.

## How

- New `bazel-workspace-walk.mts`: `findWorkspaceRoots()` — a deterministic,
  budget-bounded tree walk that prunes VCS / `node_modules` / `bazel-*` output
  symlink directories and caps the number of roots.
- `extract_bazel_to_maven.mts`: the previous single-workspace body is now
  `extractOneWorkspace()` (behavior unchanged); a thin wrapper walks the roots
  and runs it per workspace, aggregating results.
- `src/utils/glob.mts`: export the existing `IGNORED_DIRS` so the walker reuses
  the repo-wide ignore set instead of duplicating it.

## Back-compat

For the common single-workspace case (only the root is found) the output path,
return shape, and status are unchanged — verified by test.

## Scope

This is the first of a planned series that decomposes a larger reworking of the
Bazel JVM extraction pipeline into reviewable pieces. It intentionally layers on
the **existing** extraction mechanism; the metadata-`cquery` extraction swap,
the `bazel mod show_extension` discovery rewrite, and an honest
partial/complete status model are separate follow-up PRs.

### Known limitation (tracked for the completeness follow-up)

When multiple workspaces are discovered and one fails extraction while another
succeeds, the aggregate currently reports success (the per-workspace failure is
logged but not surfaced as an overall failure). The follow-up that introduces
the partial/complete status model will make a failed workspace mark the run
partial rather than silently complete.

## Testing

- `pnpm check` (lint + type-check): clean.
- New unit tests: walker behavior (prune, nested discovery, caps) and
  orchestrator integration (per-workspace extraction, mirrored manifest paths,
  single-workspace back-compat).
- Validated `findWorkspaceRoots` against a real nested-workspace repo: it
  discovers the nested `mobile/` workspace the single-workspace path missed.

---

> [!NOTE]
> **Medium Risk**
> Changes how Bazel repos are scanned and can miss nested workspaces if walk budgets cap discovery; aggregate `ok` can stay true when one workspace fails if another succeeds (noted follow-up). Core extraction per workspace is unchanged.
> 
> **Overview**
> **Bazel Maven extraction** now discovers every workspace under the scan root (directories with `MODULE.bazel`, `WORKSPACE`, or `WORKSPACE.bazel`) and runs the existing single-workspace flow once per root, instead of only treating the scan directory as one workspace.
> 
> A new **`findWorkspaceRoots`** walker performs a deterministic tree walk with caller-supplied prune rules (`IGNORED_DIRS` plus VCS/IDE dirs and `bazel-*` prefixes), no depth limit, and guards via a visited-directory budget and a 16-root cap—both log **`logger.warn`** when they truncate. **`extractBazelToMaven`** becomes a thin orchestrator around unchanged **`extractOneWorkspace`** logic; nested workspaces write manifests under **`out`** paths that mirror their relative location (root output unchanged for the single-workspace case). Results add **`manifestPaths`** while keeping **`manifestPath`** as the first root for compatibility.
> 
> **`IGNORED_DIRS`** is exported from **`glob.mts`** so the Bazel walker reuses the shared ignore list. Unit tests cover the walker and multi-workspace orchestration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 20fab7c070a4b551031b97fe121825095c27c52d. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/commands/manifest/bazel/bazel-workspace-walk.mts
+++ b/src/commands/manifest/bazel/bazel-workspace-walk.mts
@@ -108,6 +108,10 @@ export function findWorkspaceRoots(opts: FindWorkspaceRootsOptions): string[] {
       if (ignoreDirNames.has(name)) {
         continue
       }
+      // Note: `Dirent.isDirectory()` does not follow symlinks, so Bazel's
+      // `bazel-*` output symlinks are already excluded by the gate above. This
+      // prefix prune is what catches a REAL directory named `bazel-*` (and is
+      // cheap defense-in-depth for the symlink case).
       let pruned = false
       for (const prefix of ignoreDirPrefixes) {
         if (name.startsWith(prefix)) {
--- a/src/commands/manifest/bazel/bazel-workspace-walk.mts
+++ b/src/commands/manifest/bazel/bazel-workspace-walk.mts
@@ -1,0 +1,157 @@
+/**
+ * Walk the directory tree rooted at `cwd` and return every directory that
+ * looks like a Bazel workspace root — i.e. contains `MODULE.bazel`,
+ * `WORKSPACE`, or `WORKSPACE.bazel`. Real monorepos host multiple roots
+ * (e.g. `envoy/mobile/MODULE.bazel`, rules_kotlin's per-example
+ * `examples/<name>/MODULE.bazel`); the per-workspace extraction runs once
+ * per discovered root.
+ *
+ * The walker is dependency-injected with the directory-prune policy:
+ * callers pass the set of basenames and basename prefixes the walk must
+ * refuse to descend into. This module intentionally hardcodes none of
+ * the "common" prunes (`.git`, `node_modules`, …) — Bazel callers compose
+ * the codebase-wide `IGNORED_DIRS` list (`src/utils/glob.mts`) with the
+ * Bazel-specific bits (`bazel-*` output_base symlinks,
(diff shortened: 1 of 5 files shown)


simonhj:
Closing as redundant: the sub-workspace discovery (and the broader Bazel-native Maven extraction rework) this PR was the first slice of has already landed on `v1.x` via #1342. The remaining, unmerged value from this line of work is a set of correctness hardening fixes (completeness signal, committed-lockfile gating, `show_extension` failure classification, cquery timeout detection) which I'm opening as a separate PR against current `v1.x`.

PR #1373, opened 2026-06-19
feat(manifest): add `socket manifest maven` (1.1.125, Coana 15.5.5)

jfblaa:
## Summary

Adds a new `socket manifest maven` command that generates a Socket facts file (`.socket.facts.json`) from a Maven `pom.xml` project, mirroring the existing `gradle` / `scala (sbt)` facts flows. socket-cli delegates to the Coana CLI's `manifest maven` command (which owns the Maven extension that resolves the dependency graph); socket-cli only constructs the args and verifies the emitted facts file.

This ships as the **1.1.125** release (1.1.124 was taken by `--reach-retain-facts-file` on `v1.x`, since merged in): the feature plus the Coana bump it depends on.

## What's included

- **New command** `cmd-manifest-maven.mts` — facts-only (Maven is already pom-native, so there is no `--pom` path). Flags: `--bin` (defaults to `mvn` on PATH; point at `./mvnw` to use a wrapper), `--maven-opts` (e.g. `-P <profile> -s <settings.xml>`), `--include-configs` / `--exclude-configs` (Maven scopes), `--ignore-unresolved`, `--verbose`. Reads defaults from `socket.json` like the other JVM tools.
- **`convert-maven-to-facts.mts`** — thin delegate to the shared `runCoanaManifestFacts` (widened to accept `ecosystem: 'maven'` / `--maven-opts`).
- **Auto-detection** — `pom.xml` → maven in `detect-manifest-actions.mts`, wired into `socket manifest auto` (`generate_auto_manifest.mts`).
- **Setup wizard** — a Maven choice + `setupMaven` in the `socket manifest setup` configurator.
- **Types** — `maven` entry in `socket.json` manifest defaults.
- **Docs/changelog** — README section and a 1.1.125 changelog entry.
- **Coana bump** — `@coana-tech/cli` 15.5.0 → 15.5.5, which adds the `manifest maven` command this delegates to.

## Review follow-ups (mtorp)

- **Changelog version label** — relabeled the new section `1.1.124` → `1.1.125` to match `package.json`, and aligned the PR title/body.
- **`--*-opts` space-splitting** — replaced the per-tool `split(' ')` with a shared quote-aware tokenizer (`parseBuildToolOpts`) honoring single/double quotes, applied across gradle/kotlin/scala/maven + auto-manifest so the advertised `-s "my settings.xml"` case works without a maven-only divergence. Unquoted input tokenizes exactly as before; added a unit test.

## Verification

- `pnpm build:dist:src`, `pnpm check:tsc`, `pnpm check:lint`, `biome format` all clean.
- New `parse-build-tool-opts.test.mts` and `cmd-manifest-maven.test.mts`; gradle/kotlin/scala/auto suites pass.
- End-to-end smoke test against a real Maven project with the published 15.5.5 produces a complete graph (direct + transitive deps, correct `tool: "maven"` metadata).


CHANGELOG.md
```diff
@@ -4,9 +4,16 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
-## [1.1.124](https://github.com/SocketDev/socket-cli/releases/tag/v1.1.124) - 2026-06-19
+## [1.1.124](https://github.com/SocketDev/socket-cli/releases/tag/v1.1.125) - 2026-06-19
```
mtorp: **Version label mismatch.** `package.json` bumps to `1.1.125`, so this section is the 1.1.125 release — but the heading reads `[1.1.124]` and the link text (`1.1.124`) disagrees with its target tag (`v1.1.125`). It also leaves two `## [1.1.124] … - 2026-06-19` headings in the file (this one + the pre-existing one below). The PR title/body also say 1.1.124, so worth aligning all of package.json / changelog / title on one version.

```suggestion
## [1.1.125](https://github.com/SocketDev/socket-cli/releases/tag/v1.1.125) - 2026-06-19
```

src/commands/manifest/cmd-manifest-maven.mts
```diff
@@ -0,0 +1,235 @@
...
+    logger.groupEnd()
+  }
+
+  if (dryRun) {
+    logger.log(constants.DRY_RUN_BAILING_NOW)
+    return
+  }
+
+  const parsedMavenOpts = String(mavenOpts || '')
+    .split(' ')
```
mtorp: `--maven-opts` is split on every space, so a value with a spaced path — e.g. `--maven-opts="-s my settings.xml"` — becomes three separate tokens (`-s`, `my`, `settings.xml`). This matches the existing `--gradle-opts`/`--sbt-opts` parsing, so it's consistent and not a regression — flagging only because the Maven help text and changelog advertise exactly this `-s settings.xml` case. Non-blocking; if you want it to work, best fixed across gradle/sbt/maven together with a shell-style tokenizer.

mtorp (approved):
Approving — clean, low-risk addition that faithfully mirrors the existing Gradle/sbt facts delegation: type-safe flag/`socket.json` plumbing, an argv-based (no-shell) spawn path, and good rationale comments. CI is green across the board.

Two non-blocking notes inline:
- **CHANGELOG version label** — the new section is the `1.1.125` release (per `package.json`) but is labeled `1.1.124`; suggestion attached. Worth reconciling with the PR title/body too (they say 1.1.124).
- **`--maven-opts` space-splitting** — a pre-existing, consistent limitation shared with `--gradle-opts`/`--sbt-opts`; flagged only because the Maven help text advertises the spaced-path case.

CHANGELOG.md
jfblaa: Fixed in aee350906. Relabeled the heading `1.1.124` → `1.1.125` (your exact suggestion; the link target was already `v1.1.125`), so there's now a single `1.1.124` section (the pre-existing `--reach-retain-facts-file` release merged in from `v1.x`) and a distinct `1.1.125` section for this change. Also aligned the PR title and body on `1.1.125` to match `package.json`.

src/commands/manifest/cmd-manifest-maven.mts
jfblaa: Fixed in aee350906, and done across all three tools as you suggested rather than as a maven-only change. Added a shared quote-aware tokenizer `parseBuildToolOpts` (`src/commands/manifest/parse-build-tool-opts.mts`) that splits on whitespace but honors single/double quotes (and is intra-token aware, so `-Dprop="a b"` → `-Dprop=a b`). Replaced the per-tool `split(' ')` at all manifest opts sites: `--gradle-opts`, `--sbt-opts`, `--maven-opts` in the cmd files plus the three `generate_auto_manifest.mts` sites. Now `--maven-opts="-s 'my settings.xml'"` survives as one token. Unquoted input tokenizes exactly as before (no behavior change for existing usage), and there's a unit test (`parse-build-tool-opts.test.mts`) covering empty input, plain splitting, whitespace runs, and single/double/intra-token quoting.

PR #1377, opened 2026-06-23
fix(glob): bound gitignore matching memory to prevent scan OOM

simonhj:
`socket fix` and `socket scan` abort with `FATAL ERROR: CALL_AND_RETRY_LAST … heap out of memory` on large monorepos that contain many nested `.gitignore` files.

## Cause

`globWithGitIgnore` discovers every nested `.gitignore`, unions all their translated patterns into one set, and handed that entire set to fast-glob's native `ignore` option. fast-glob re-compiles and re-tests its whole ignore array inside every directory scan, so a union of tens of thousands of patterns exhausts V8 code space and aborts the process. Raising `--max-old-space-size` does not help — the allocation is regex executable code, not the data heap.

## Fix

Match the high-cardinality gitignore set through a single reused `ignore` instance (which compiles each rule once and memoizes it) applied per streamed entry, and hand fast-glob only the small bounded set it needs to prune directories during the walk. The negated-pattern path already worked this way; this unifies both paths and removes the asymmetry that left the common, non-negated case crashing.

Two parity details versus fast-glob's native ignore matching are preserved:

- Case sensitivity tracks `caseSensitiveMatch` (default case-sensitive, matching git) rather than the `ignore` package's case-insensitive default, so `dist/` no longer also ignores a differently-cased `Dist/`.
- The cwd-relative path is normalized to POSIX separators before matching, so it still matches the forward-slash-anchored patterns on Windows.

## Tests

- A regression test builds a 100k-pattern nested-`.gitignore` tree and asserts the walk completes with the correct manifests; the pre-fix path exhausts a constrained worker heap at that count.
- A case-sensitivity test asserts `dist/` ignores `dist/` but leaves `Dist/` alone.
- The existing `glob` suite stays green.

---

> [!NOTE]
> **Medium Risk**
> Core file-discovery path for scan/fix; behavior is intentionally unified (always streaming) with parity fixes for case and Windows paths, but regressions in ignore semantics could miss or over-exclude manifests.
> 
> **Overview**
> Fixes **`socket scan` / `socket fix` heap OOM** on monorepos with many nested `.gitignore` files by changing how **`globWithGitIgnore`** applies the unioned ignore set.
> 
> Instead of passing tens of thousands of gitignore-derived patterns into **fast-glob**’s `ignore` option (which recompiles the full list on every directory scan), the walk now streams paths, applies a **single reused `ignore` instance** per entry, and gives fast-glob only the small **`defaultIgnore` + CLI `additionalIgnores`** set for directory pruning. The old fast path and negated-vs-non-negated split are removed so all scans use this streaming + `ig` path.
> 
> **Parity:** ignore matching respects **case-sensitive** git/fast-glob defaults via `ignorecase`, and cwd-relative paths are **`normalizePath`**’d before matching so Windows backslashes still hit forward-slash patterns.
> 
> Adds a **100k-pattern regression test** (`glob-oom.test.mts`) and a **case-sensitivity** test; **`path-resolve`** comments clarify that streaming filters bound result memory, not ignore-pattern memory.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 90890ee5d4a4169eb32b1ea70be4a518bc9f284c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/utils/glob-oom.test.mts
+++ b/src/utils/glob-oom.test.mts
@@ -1,0 +1,72 @@
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import path from 'node:path'
+
+import { describe, expect, it } from 'vitest'
+
+import { normalizePath } from '@socketsecurity/registry/lib/path'
+
+import { globWithGitIgnore } from './glob.mts'
+
+// Defined at module scope to satisfy linting rules.
+function filterJsonFiles(filepath: string): boolean {
+  return filepath.endsWith('.json')
+}
+
+// This suite lives in its own file, with no mock-fs node_modules preload, so the
+// large ignore set it builds is the only significant allocation in the worker.
+describe('globWithGitIgnore() large monorepo memory', () => {
+  // Regression: `socket fix` / `socket scan` aborted with
+  // `FATAL ERROR: CALL_AND_RETRY_LAST … heap out of memory` (SIGABRT) on large
+  // monorepos. globWithGitIgnore discovers every nested .gitignore and unions
+  // their patterns; handing that whole set to fast-glob's `ignore` option made
+  // fast-glob re-compile and re-test the entire array inside every directory
+  // scan, so tens of thousands of patterns exhausted V8 code space. Routing the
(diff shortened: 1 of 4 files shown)


simonhj:
Closing, was a red herring.

PR #1379, opened 2026-06-25
fix(scan): exclude Python virtual environments from manifest collection + bump Coana CLI to 15.5.10 (1.1.128)

mtorp:
This PR ships in release **1.1.128** and contains two changes.

### Problem

When the CLI collects manifest files for a scan (`socket scan`, `socket scan --reach`, and `socket fix`), it walks the target tree recursively and excludes a fixed set of directories (`node_modules`, `.git`, `.yarn`, …) plus anything matched by `.gitignore` / `socket.yml` rules. **Python virtual environments are not excluded.** A venv's `lib/.../site-packages` tree holds thousands of installed packages, each with its own `setup.py`, `pyproject.toml`, `requirements.txt`, etc. Those get picked up as manifests — bloating scans, polluting results with dependency-of-dependency manifests, and wasting a full-tree walk into a directory that should never be scanned.

Venvs are usually named `.venv` or `venv`, but the name is arbitrary. The reliable, name-independent signal is the **`pyvenv.cfg`** file that Python's stdlib `venv` (PEP 405) and `virtualenv` ≥ 20 always write at the environment root — nothing else creates that file, so there are no false positives.

### Change

In `src/utils/glob.mts` (the single path all three commands funnel through, `getPackageFilesForScan` → `globWithGitIgnore`):

- Add `.venv` to `IGNORED_DIRS` for a cheap traversal-prune of the conventional directory name (also catches pre-2020 virtualenvs that lack a `pyvenv.cfg`).
- Detect arbitrarily-named venvs by their `pyvenv.cfg` marker. Discovery is **folded into the existing `.gitignore` discovery walk**, so it adds no extra full-tree traversal. Each detected venv root contributes a `<dir>/**` ignore pattern to the shared `ignores` set, which every downstream glob path (fast, filter-streaming, and negated-pattern) already honors.

Bare `venv` / `env` are intentionally **not** name-excluded, to avoid skipping a legitimately-named non-venv directory that holds real manifests; the `pyvenv.cfg` check covers them correctly (a real venv has the marker; a non-venv directory does not).

Out of scope: conda environments (no `pyvenv.cfg`) and the separate Bazel manifest walker.

### Tests

Added cases to `src/utils/glob.test.mts`:
- Excludes an arbitrarily-named venv (`myenv/`) detected via `pyvenv.cfg`, keeping the project's own root manifest.
- Excludes `.venv` by name.
- **Keeps** a non-venv directory named `venv/` that has no `pyvenv.cfg` (regression guard against over-exclusion).
- Confirms the exclusion also prunes through the streaming-filter path that manifest scanning actually uses.

Verification: `glob.test.mts` 18/18 and `path-resolve.test.mts` 14/14 pass; `pnpm check` (lint + tsc) is green.

## 2. Bump Coana CLI to 15.5.10

Upgrades `@coana-tech/cli` from `15.5.9` to `15.5.10` (`package.json` + `pnpm-lock.yaml`). For details on what's included in this Coana release, see the [Coana Changelogs](https://docs.coana.tech/changelogs).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [1.1.128](https://github.com/SocketDev/socket-cli/releases/tag/v1.1.128) - 2026-06-25
+
+### Fixed
+- Scans now skip Python virtual environments when collecting manifest files. Folders named `.venv`, and any folder containing a `pyvenv.cfg` marker (covering `venv`, `env`, and custom-named environments), are excluded — so `socket scan`, reachability, and `socket fix` stay focused on your project's own manifests instead of the thousands installed inside a virtualenv.
+
 ## [1.1.127](https://github.com/SocketDev/socket-cli/releases/tag/v1.1.127) - 2026-06-24
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [1.1.128](https://github.com/SocketDev/socket-cli/releases/tag/v1.1.128) - 2026-06-25
 
+### Changed
+- Updated the Coana CLI to v `15.5.10`.
+
 ### Fixed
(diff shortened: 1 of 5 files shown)


BarrensZeppelin (approved):
LGTM ✅

barslev (approved):

PR #1385, opened 2026-06-30
feat(manifest): generate JVM Socket facts natively + single-run reachability sidecar (1.1.132, Coana 15.6.3)

jfblaa:
## Summary

Moves JVM build-tool resolution (Gradle, sbt, Maven) into socket-cli, so `socket manifest` and `--auto-manifest` generate Socket facts natively. On `socket scan create --reach --auto-manifest`, the build tool runs **once** and produces both the SBOM and a resolved-paths sidecar for coana's reachability — instead of coana re-resolving the build at reach time, which diverged for dynamically-versioned projects (git versions, CI build numbers, timestamps) and surfaced as spurious "failed to install" / stale-manifest failures.

## Related

- https://linear.app/socketdev/issue/REA-613
- https://linear.app/socketdev/issue/REA-607 — original reported issue
- https://github.com/coana-tech/coana-package-manager/pull/2292 — coana side

## Release gating

⚠️ **Draft — do not merge yet.** Gated on the coana release that ships `coana run --compute-artifacts-sidecar`. Before merge: bump `@coana-tech/cli` to that version and fill the `CHANGELOG.md` placeholder (coana version + date).

## Test plan

- `pnpm check` (lint + tsc) and unit tests green.
- Validated on Gradle (including a dynamic git-version build) and a multi-module Maven project; full `--reach --auto-manifest` end-to-end is pending the coana release.

---

> [!NOTE]
> **Medium Risk**
> Large new surface area (build-tool scripts, graph assembly, publish-time Maven jar) affects manifest and reachability correctness; behavior changes for all Gradle/sbt/Maven users but is covered by unit tests and on-demand compat matrices.
> 
> **Overview**
> **Gradle, sbt, and Maven Socket facts are generated inside socket-cli** instead of delegating to `coana manifest`. Bundled Gradle init script, sbt plugin, and a Maven core extension emit tab-separated **records**; TypeScript (`runManifestScript` → `assembleFacts`) assembles `.socket.facts.json` and handles resolution failure reporting per ecosystem.
> 
> For **`--auto-manifest` / reachability**, the same build run can use **`--with-files`** to collect resolved jar paths and module source/output dirs into a **`ResolvedPathsSidecar`** (merged across JVM roots), so reachability reuses one resolution instead of Coana re-running the build (fixes dynamic-version / install-error mismatches).
> 
[...]

PR #1386, opened 2026-06-30
chore(ci): remove the unused maven-extension-jar workflow

jfblaa:
## Summary

Removes the `maven-extension-jar` workflow added in #1385. It's redundant:

- The Maven extension jar is built and bundled at **release** by `provenance.yml`, which now **fails closed** if the jar is missing (no jarless package can ship).
- The build-tool version-compat matrix is run **locally on demand** via `src/commands/manifest/scripts/test/run-compat.sh`.

So the standalone PR-time jar-build check isn't needed. The trade-off: a maven-extension build break is caught at release time (the release fails) rather than on the PR — acceptable given the fail-closed release build and local testing.

---

> [!NOTE]
> **Low Risk**
> CI-only removal with no runtime or release-script changes; jar build and fail-closed bundling remain in the release workflow.
> 
> **Overview**
> Deletes the **`.github/workflows/maven-extension-jar.yml`** workflow that built the Maven manifest extension jar and ran a Maven 3.9.9 smoke test on PRs and `workflow_dispatch`.
> 
> **PR-time coverage is dropped** in favor of existing paths: **`provenance.yml`** still runs `pnpm run build:maven-extension` before publish and **fails closed** if the jar is missing, and the version-compat matrix is exercised locally via **`run-compat.sh`**. A broken extension build would surface at **release** rather than on every PR touching the extension.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ce178219bb68131283b22399229f11453c9512ca. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jfblaa:
Folded into #1390, which removes this workflow alongside the direct-invocation build fix.

PR #1398, opened 2026-07-02
feat(manifest): unify Gradle config-level resolution transparency (REA-519)

jfblaa:
## What & why

Follow-up to REA-516. Two Gradle resolution paths sat in tension with the "reveal what we couldn't resolve" policy:

1. A whole configuration whose resolution **throws** (as opposed to a single unresolved dependency within it) was caught and logged to stdout, but never surfaced — it still counted as `scanned`, so the user was never told we couldn't scan it.
2. The `--include-configs` help still advertised a default AGP-instrumented-test exclusion that the native-facts rewrite had already removed.

## Changes

**REA-519 — config-level transparency**
- The per-config `try/catch` now records the config as a new `unscannable` fact (build-tool's deepest-cause message) instead of swallowing it. The build script still exits cleanly; the TS layer owns reporting.
- `unscannable` configs are classified by the **same cause rules as per-dependency failures**: variant ambiguity stays lenient (a one-line notice), every other cause is fail-closed (blocking, overridable by `--ignore-unresolved`).
- Un-scannable configs are excluded from "Resolution succeeded in"; full messages show under `--verbose`.
- A config-level failure no longer trips the crashed-build path (which stays reserved for a build that produced nothing at all).
- Dropped the stale "default excludes AGP instrumented-test classpaths" help text (+ snapshots).

**REA-516 — classifier clarification (first commit)**
- Recognize Gradle's alternate zero-compatible-variant phrasings (`unable to find a matching variant`, `no compatible variant`) so they render under the accurate "No compatible variant" header instead of falling through to "Other". Severity unchanged (both blocking).

## Empirical validation

Validated on real throwaway Gradle projects (AGP 8.13.2 / Gradle 9.2.1 / JDK 21):
- **AGP `*AndroidTest*` configs resolve cleanly** from the init script (41 nodes/config, no ambiguity, no throws) — the hypothesized "init-script can't supply consumer attributes → ambiguity" does not occur, because AGP sets the attributes on the configuration itself. Removing the name-based exclusion was safe.
- Gradle's `lenientConfiguration` **swallows all dependency-graph failures** (not-found, capability conflict, version conflict, variant ambiguity) into per-dep records. The `unscannable` path therefore fires only on **non-graph** exceptions (resolution-rule throw, artifact-transform failure, broken repo config, …) — a rare but real safety net.
- The full `unscannable` pipeline was exercised end-to-end (one config throwing, others resolving): default → exit 1 + no facts written; `--ignore-unresolved` → exit 0 + partial facts written.

## Testing
- `pnpm check:tsc` clean; targeted unit tests green (render/assemble/gradle/kotlin help snapshots).
- Verified end-to-end via `socket manifest gradle --facts` against the test projects.

---

> [!NOTE]
> **Medium Risk**
> Changes exit-code and facts-write behavior when Gradle configurations throw during resolution (partial SBOM vs hard fail), which affects manifest/scan workflows on Android and complex Gradle builds; logic is covered by new render tests and mirrors existing per-dep failure policy.
> 
> **Overview**
> Gradle manifest facts generation now **records whole-configuration resolution failures** instead of swallowing them in the init script. Thrown configs emit **`unscannable`** records through the line protocol; the TypeScript layer classifies each message with the **same rules as per-dependency failures** (variant ambiguity → non-blocking notice; other causes → blocking, honor **`--ignore-unresolved`**). Reports distinguish “could not scan N configuration(s)” from unresolved deps, omit unscannable configs from “Resolution succeeded in”, and include them in **`--verbose`** output. **Crashed-build** detection no longer treats unscannable-only outcomes as a total build failure.
> 
> **`--include-configs`** help for **`manifest gradle`** / **`manifest kotlin`** no longer claims a default AGP instrumented-test exclusion (default is every resolvable configuration). Gradle’s **“unable to find a matching variant”** / **“no compatible variant”** wording maps to the existing **no-matching-variant** blocking category.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0c65404cc0d2c5b9f9aaebe60b1c264cdf3df43a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/commands/manifest/scripts/resolution-report-render.mts
+++ b/src/commands/manifest/scripts/resolution-report-render.mts
@@ -193,7 +193,11 @@ export function renderResolutionReport(
       }
     }
     if (blockingUnscannable.length) {
-      out.push('')
+      // Separate from the per-dep block above, but only if there is one — otherwise
+      // the summary would lead with a blank line (a dangling ✗ under logger.fail).
+      if (out.length) {
+        out.push('')
+      }
       out.push(
         opts.ignoreUnresolved
           ? `Ignored ${blockingUnscannable.length} configuration(s) that could not be scanned:`
@@ -243,8 +247,8 @@ export function renderResolutionReport(
     const configCount = new Set(infos.flatMap(i => [...i.configs])).size
     notices.push(spec.notice(name, infos.length, configCount))
   }
-  // Ambiguity-driven config-throws (e.g. AGP *AndroidTest*) are surfaced, not failed:
-  // deliberately lenient, matching the per-dep variant-ambiguity policy.
+  // A config-level throw whose cause classifies as variant ambiguity is surfaced, not failed —
+  // matching the deliberately-lenient per-dep variant-ambiguity policy.
   if (nonBlockingUnscannable.length) {
     const n = new Set(nonBlockingUnscannable.map(u => u.config)).size
     notices.push(
--- a/src/commands/manifest/scripts/resolution-report-render.mts
+++ b/src/commands/manifest/scripts/resolution-report-render.mts
@@ -3,7 +3,10 @@ import { SBT_DIALECT } from './resolution-report-ivy.mts'
 import { MAVEN_DIALECT } from './resolution-report-maven.mts'
 
 import type { BuildTool } from './build-tool.mts'
-import type { ResolutionFailure } from './resolution-report.mts'
+import type {
+  ResolutionFailure,
+  UnscannableConfig,
+} from './resolution-report.mts'
 
 // Recognized from the build tool's message; drives wording AND whether the kind
 // is blocking. An unrecognized message degrades to 'other' (blocking) — safe.
@@ -87,9 +90,14 @@ export function renderResolutionReport(
   failures: ResolutionFailure[],
   scannedConfigs: string[],
   dialect: ResolutionDialect,
-  opts: { ignoreUnresolved?: boolean | undefined } = {},
+  opts: {
+    ignoreUnresolved?: boolean | undefined
+    unscannable?: UnscannableConfig[] | undefined
+  } = {},
 ): RenderedResolutionReport {
   const name = dialect.label
+  const unscannable = opts.unscannable ?? []
+  const unscannableConfigs = new Set(unscannable.map(u => u.config))
   const specOf = new Map(dialect.categories.map(c => [c.key, c]))
   const isBlocking = (cat: FailureCategory): boolean =>
     specOf.get(cat)?.blocking ?? true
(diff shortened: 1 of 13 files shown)


src/commands/manifest/scripts/resolution-report-render.mts
```diff
@@ -141,27 +165,50 @@ export function renderResolutionReport(
...
+        }
+        if (infos.length > RESOLUTION_REPORT_ARTIFACT_LIMIT) {
+          out.push(
+            `    … and ${infos.length - RESOLUTION_REPORT_ARTIFACT_LIMIT} more`,
+          )
+        }
+      }
+    }
+    if (blockingUnscannable.length) {
       out.push('')
```
mtorp: Cosmetic nit: when there are no per-dep blocking failures (`blockingCount === 0`), this unconditional separator makes the summary start with a blank line, so `logger.fail` renders a dangling ✗ on an empty line. Guarding it with `if (out.length)` would fix that.

src/commands/manifest/scripts/socket-facts.init.gradle
```diff
@@ -377,7 +379,9 @@ allprojects { project ->
             ]
           }
         } catch (Exception e) {
-          println "[socket-facts] skipping ${project.path}:${cfg.name}: ${e.message?.readLines()?.first()}"
+          // Record rather than swallow (e.g. AGP *AndroidTest* configs the init script can't resolve);
```
mtorp: Nit: this comment (and the matching one in `resolution-report-render.mts`) cites AGP `*AndroidTest*` configs as the example, but per your own validation in the PR description those resolve cleanly from the init script, and this path only fires on non-graph exceptions (resolution-rule throws, artifact-transform failures, broken repo config, …). Might be worth updating the examples so future readers don't assume AndroidTest configs land here.

mtorp (approved):
LGTM — traced the `unscannable` pipeline end-to-end (init script → records → assemble → render → exit policy), and typecheck + the render/help-snapshot tests pass locally. Two cosmetic nits inline, neither blocking.

src/commands/manifest/scripts/resolution-report-render.mts
jfblaa: Good catch — fixed in ed3675042. Guarded the separator with `if (out.length)` so the summary no longer leads with a blank line when there are only blocking unscannable configs (no per-dep failures). Added a regression assertion (`summary` must not start with `\n`) to the blocking-config-throw test.

src/commands/manifest/scripts/socket-facts.init.gradle
jfblaa: Agreed — the example was misleading. Fixed in ed3675042. Both comments (this one in `socket-facts.init.gradle` and the matching one in `resolution-report-render.mts`) now describe the path as config-LEVEL failures — a resolutionStrategy/beforeResolve rule that throws, an artifact-transform failure, a broken repository — rather than AGP `*AndroidTest*` configs, which resolve cleanly per the validation.

PR #1400, opened 2026-07-02
feat(manifest): case-sensitive config-name globs with character classes (REA-621)

jfblaa:
## Problem

The `--include-configs` / `--exclude-configs` glob matchers in all three JVM facts producers — the Gradle init script (Groovy), the sbt plugin (Scala), and the Maven extension (Java) — compiled their patterns with `Pattern.CASE_INSENSITIVE`. That contradicts the CLI help ("case-sensitive") and the capitalized examples (`*CompileClasspath,*RuntimeClasspath`). Confirmed: `--include-configs '*classpath'` (lowercase) matched `compileClasspath` etc.

## Changes

1. **Case-sensitive matching** — dropped `CASE_INSENSITIVE` in all three `globToRegex` implementations, matching the documented behavior. (The old code even had a comment justifying the insensitivity on a false premise — `*Classpath` matches camelCase `compileClasspath` case-sensitively just fine.)
2. **Character-class support** — `[...]` now works: enumerations (`[cC]`), ranges (`[a-z]`), and `[!..]` negation. Under case-sensitive matching this lets a pattern span the camelCase boundary, e.g. `*[Tt]est*` matches both `testCompileClasspath` and `androidTestCompileClasspath` — impossible with `*`/`?` alone.
3. **Robustness** — the class body is hardened (escape `\`/`[`/`&`) so exotic input (POSIX classes, `&&` intersection) degrades to a literal match, and `Pattern.compile` is wrapped so a malformed glob (e.g. reversed range `[z-a]`) falls back to a literal match instead of throwing and failing the build.
4. Help text for gradle/kotlin/sbt/maven mentions `[...]`; fixed a stale "case-insensitive" javadoc on `SocketSupport.parsePatterns`.

## Why a custom converter (not a library / `PathMatcher`)

Considered `java.nio.file.PathMatcher` `glob:` and TS-side conversion (picomatch). Both were rejected:
- **`PathMatcher`** is case-sensitive on Unix but **case-insensitive on Windows** (JDK `Globs.toWindowsRegexPattern` adds `CASE_INSENSITIVE`), and `Path.of` is **Java 11+** — this code must run on **Java 8→newest, Gradle 1→newest, sbt 0.13→newest, Maven 3.3→newest**. `java.util.regex.Pattern` + string ops are stable since Java 1.4 and behave identically across that whole matrix.
- **picomatch/micromatch** emit **JS-flavored, path-oriented** regex (dotfile guards, `[^/]`, trailing `/?`) and even mangle `[Tt]` into `(?:\[Tt\]|[Tt])` — not portable to Java's regex engine, and passing a regex would change the producers' glob contract and add escaping fragility.

An independent review agent reached the same conclusion (custom converter), flagging the hardening/compile-guard and stale comment fixed here.

## Testing

- End-to-end against a real Gradle project: `*classpath` (lowercase) now matches nothing; `*Classpath` matches all four classpath configs; `*[Tt]est*` matches only the test configs; malformed `[z-a]` exits 0 (literal fallback, no crash).
- Maven extension jar rebuilds from source (gitignored, built in CI).
- Help snapshots updated; `pnpm check:tsc` clean; 12/12 manifest help tests pass.

**Follow-up (noted, not in this PR):** a shared cross-language conformance fixture would guard against the three hand-written converters drifting.

Closes REA-621. Part of the same Unreleased version as PRs #1398 (REA-519) and #1399 (REA-622). Note: the gradle/kotlin `--include-configs` help line also conflicts trivially with #1398 (which removes stale AGP text from the same string) — resolve by keeping both edits.

---

> [!NOTE]
> **Medium Risk**
> Matching semantics change which build configurations/scopes are resolved into facts/SBOMs; users with lowercase or case-loose patterns may see different dependency coverage until patterns are fixed.
> 
> **Overview**
> **`--include-configs` / `--exclude-configs`** on JVM manifest commands now match configuration/scope names **case-sensitively**, aligning runtime behavior with the documented CLI help. Patterns that relied on case folding (e.g. `*classpath` matching `compileClasspath`) will stop matching unless updated.
> 
> The Gradle init script, sbt facts plugin, and Maven extension **`globToRegex`** implementations are updated in parallel: they drop `CASE_INSENSITIVE`, add **`[...]`** glob classes (including `[!…]` negation), harden class bodies against regex tricks, and **fall back to a literal match** when compilation fails so bad patterns do not crash the build.
> 
> CLI flag descriptions and help snapshots for **`socket manifest gradle`**, **`kotlin`**, **`scala`**, and **`maven`** now mention `[...]` wildcards; **CHANGELOG** records the behavior change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 33850b408521573ac2b51c12e5c3b2fb0be976a6. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,24 +4,14 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
-## [Unreleased]
+## [1.1.136](https://github.com/SocketDev/socket-cli/releases/tag/v1.1.136) - 2026-07-02
 
 ### Changed
-- `socket manifest gradle --facts` no longer silently skips a Gradle
-  configuration it can't resolve. Such configurations are now reported and stop
-  the run unless you pass `--ignore-unresolved`, so an incomplete scan can't slip
-  by unnoticed. Benign variant-selection ambiguity stays a one-line notice.
-- `--include-configs` / `--exclude-configs` patterns for `socket manifest
-  gradle`, `kotlin`, `scala`, and `maven` are now matched case-sensitively (as
-  documented) and support `[...]` character classes — e.g. `*[Tt]est*` matches
-  both `testCompileClasspath` and `androidTestCompileClasspath`.
+- `socket manifest gradle --facts` now reports Gradle configurations it can't resolve instead of silently skipping them, failing the run unless you pass `--ignore-unresolved` so an incomplete scan can't slip by. Benign variant-selection ambiguity stays a one-line notice.
+- Config filters (`--include-configs` / `--exclude-configs`) for `socket manifest gradle`, `kotlin`, `scala`, and `maven` are now case-sensitive, as documented, and support `[...]` character classes — e.g. `*[Tt]est*` matches both `testCompileClasspath` and `androidTestCompileClasspath`.
 
 ### Fixed
-- `socket manifest auto` and `scan create --auto-manifest` now detect a Gradle
-  project by its build files (`build.gradle`/`.kts` or `settings.gradle`/`.kts`)
-  instead of requiring a `gradlew` wrapper, so a project that builds with
-  `gradle` on your PATH — including a settings-only multi-module root — is no
-  longer skipped.
+- `socket manifest auto` and `socket scan create --auto-manifest` now detect a Gradle project by its build files (`build.gradle`/`.kts` or `settings.gradle`/`.kts`) instead of requiring a `gradlew` wrapper, so wrapper-less projects — including settings-only multi-module roots — are no longer skipped.
 
 ## [1.1.135](https://github.com/SocketDev/socket-cli/releases/tag/v1.1.135) - 2026-07-01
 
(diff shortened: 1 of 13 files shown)


mtorp (approved):
Approving — the three-language `globToRegex` port (Groovy/Scala/Java) is faithful and consistent, the escape set is complete, and malformed globs degrade safely to a literal match via the `PatternSyntaxException` → `Pattern.quote` fallback. Case-sensitive matching plus `[...]` classes matches the documented help.

One nit for the CHANGELOG (non-blocking): the Unreleased entry lists "`socket manifest` gradle, kotlin, sbt, and maven", but there is no `socket manifest sbt` subcommand — the sbt producer is invoked via `socket manifest scala`. Suggest `scala` (or `scala (sbt)`) so users can find the command by name.

jfblaa:
Good catch on the CHANGELOG — fixed in 28aae8c9. Changed the sbt entry to `scala` (the actual subcommand: `socket manifest scala`), so the list now reads `gradle`, `kotlin`, `scala`, and `maven`.

Also merged latest `v1.x` (now includes #1398 and #1399) and resolved the conflicts: the CHANGELOG Unreleased section keeps all three entries, and the `--include-configs` help line + snapshots keep both edits — the `[...]` wildcard phrasing from this PR and the AGP-instrumented-test-default removal from #1398.

PR #1404, opened 2026-07-07
feat(manifest): socket manifest dotnet + centralized config-name globs (REA-627, REA-628)

jfblaa:
Closes REA-627. Closes REA-628.

> [!WARNING]
> **Draft — do not merge until coana PR (REA-626) is released and the pinned coana version is bumped in this PR.** The reachability sidecar emits nuget-tagged entries that the currently-released coana rejects (strict schema); a polyglot .NET+JVM `--reach` scan would break until coana ships REA-626. The SBOM-only path (no `--reach`) is coana-independent.

## What

Brings .NET/NuGet to the Socket facts pipeline, at parity with the JVM tools (the .NET analog of REA-613). Previously `nuget` was only a package-scoring ecosystem — no manifest generation, no auto-detection, no reachability; the only path was a manual `socket manifest cdxgen -t dotnet`.

### `socket manifest dotnet` (REA-627)
- New command + auto-manifest detection (top-level `.sln`/`.slnx`/`.csproj`/`.fsproj`/`.vbproj`) + setup-wizard entry; `socket.json` `defaults.manifest.dotnet.*`.
- **Bundled C# tool** (`scripts/dotnet-tool/`): one MSBuild session — evaluate → in-process restore → read `project.assets.json` via `NuGet.ProjectModel` — emitting the shared records TSV.
  - Ships **no** NuGet/MSBuild runtime assemblies: compile-low/run-high against the locator-selected SDK (net6 floor, `RollForward LatestMajor`), so it works across installed SDKs and avoids the ref/def assembly clash that breaks multi-SDK hosts.
  - Restore forces `TreatWarningsAsErrors=false` so `NU19xx` security-advisory warnings don't abort the run.
  - **packages.config** supported (flat pinned closure + `developmentDependency`→dev + HintPath DLLs; downloads missing pinned artifacts via the project's configured feeds/credential providers).
  - **Fail-closed**: emits `failure`/`unscannable` records that raise the CLI exit code; every emitted artifact path is guaranteed to exist.
- Flags: `--target-frameworks` / `--exclude-target-frameworks` (TFM globs; RID targets match under their base TFM); `--dotnet-opts` (MSBuild `-p:` props applied to the whole session).
- Reachability sidecar emits nuget-tagged `ResolvedComponent`s (consumed by coana REA-626).
- Resolution summary reports scanned target frameworks across N projects (per-project attribution under `--verbose`).
- CI leg builds the C# tool; provenance build bundles it like the Maven extension jar.

### Centralized config-name glob compilation (REA-628)
- Compile `--include-configs`/`--exclude-configs` globs → anchored regex sources once in `config-glob.mts`; the gradle/sbt/maven scripts and the dotnet tool all consume the compiled patterns. Removes the per-language `globToRegex` (Groovy/Scala/Java) and adds a cross-language vector test suite. Rides in this PR because the .NET tool is a fourth consumer of the shared patterns.

## Testing
Unit tests (config-glob vectors, dotnet records→assemble, detection, auto-manifest branch, command snapshots); accuracy sweep vs. native-restore ground truth across 17 real .NET repos (~99.6% avg recall, exact real-package versions, packages.config 100%). `pnpm check` clean.

## Follow-ups before un-drafting
- [ ] coana REA-626 merged + released
- [ ] bump pinned coana version in this PR
- [ ] CHANGELOG entry

### DIFF
--- a/.config/rollup.dist.config.mjs
+++ b/.config/rollup.dist.config.mjs
@@ -120,6 +120,20 @@ async function copyManifestScripts() {
         ' for a published build. Build it first: pnpm run build:maven-extension',
     )
   }
+  // Same contract for the dotnet tool: compiled by dotnet-tool/build-tool.sh,
+  // tolerated when absent in local dev, required for published builds.
+  const dotnetPublishDir = path.join(srcDir, 'dotnet-tool', 'publish')
+  if (existsSync(path.join(dotnetPublishDir, 'socket-facts-dotnet.dll'))) {
+    await fs.cp(dotnetPublishDir, path.join(destDir, 'dotnet-tool'), {
+      recursive: true,
+    })
+  } else if (constants.ENV[INLINED_SOCKET_CLI_PUBLISHED_BUILD]) {
+    throw new Error(
+      'socket-facts-dotnet tool not found at ' +
+        dotnetPublishDir +
+        ' for a published build. Build it first: pnpm run build:dotnet-tool',
+    )
+  }
 }
 
 async function copyBashCompletion() {
--- a/.github/workflows/provenance.yml
+++ b/.github/workflows/provenance.yml
@@ -220,6 +220,13 @@ jobs:
           fi
           bash src/commands/manifest/scripts/maven-extension/build-jar.sh
 
+      # Same bundling contract as the Maven jar: the socket-facts-dotnet tool is
+      # never committed and ships only in the published package. Invoked via bash
+      # directly for the same Socket Firewall shim reason as above; uses the .NET
+      # SDK pre-installed on the runner image.
+      - name: Build dotnet manifest tool
+        run: bash src/commands/manifest/scripts/dotnet-tool/build-tool.sh
+
       - run: INLINED_SOCKET_CLI_PUBLISHED_BUILD=1 pnpm run build:dist
       - name: Publish socket
         id: publish_socket
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "build:dist": "pnpm build:dist:src && pnpm build:dist:types",
     "build:dist:src": "run-p -c clean:dist clean:external && dotenvx -q run -f .env.local -- rollup -c .config/rollup.dist.config.mjs",
     "build:dist:types": "pnpm clean:dist:types && tsgo --project tsconfig.dts.json",
+    "build:dotnet-tool": "bash src/commands/manifest/scripts/dotnet-tool/build-tool.sh",
     "build:maven-extension": "bash src/commands/manifest/scripts/maven-extension/build-jar.sh",
     "build:sea": "node src/sea/build-sea.mts",
     "build:sea:internal:bootstrap": "rollup -c .config/rollup.sea.config.mjs",
--- a/src/commands/manifest/cmd-manifest-dotnet.mts
+++ b/src/commands/manifest/cmd-manifest-dotnet.mts
@@ -1,0 +1,249 @@
+import path from 'node:path'
+
+import { debugFn } from '@socketsecurity/registry/lib/debug'
+import { logger } from '@socketsecurity/registry/lib/logger'
+
+import { convertDotnetToFacts } from './convert-dotnet-to-facts.mts'
+import { parseBuildToolOpts } from './parse-build-tool-opts.mts'
+import constants, { SOCKET_JSON } from '../../constants.mts'
+import { commonFlags } from '../../flags.mts'
+import { checkCommandInput } from '../../utils/check-input.mts'
+import { getOutputKind } from '../../utils/get-output-kind.mts'
+import { meowOrExit } from '../../utils/meow-with-subcommands.mts'
+import { getFlagListOutput } from '../../utils/output-formatting.mts'
+import { readOrDefaultSocketJson } from '../../utils/socket-json.mts'
+
+import type {
+  CliCommandConfig,
+  CliCommandContext,
+} from '../../utils/meow-with-subcommands.mts'
+
+const config: CliCommandConfig = {
+  commandName: 'dotnet',
+  description:
+    '[beta] Generate a Socket facts file for a .NET (C#/F#/VB) project',
+  hidden: false,
+  flags: {
+    ...commonFlags,
+    bin: {
+      type: 'string',
+      description:
+        'Location of the dotnet binary to use, default: dotnet on PATH',
+    },
+    targetFrameworks: {
+      type: 'string',
+      description:
+        'Comma-separated glob patterns matched against target framework names (case-sensitive; `*`, `?`, and `[...]` wildcards). Only target frameworks matching at least one pattern are included, e.g. `net8.0` or `net*`. Default: every restored target framework',
+    },
+    excludeTargetFrameworks: {
+      type: 'string',
+      description:
+        'Comma-separated glob patterns; target frameworks matching any pattern are skipped (applied after --target-frameworks)',
+    },
+    ignoreUnresolved: {
+      type: 'boolean',
+      description:
+        'Warn on restore/resolution failures instead of failing the run (unresolved deps are not emitted to the facts file)',
+    },
+    dotnetOpts: {
+      type: 'string',
+      description:
+        'MSBuild property tokens (`-p:Key=Value`) applied to the whole facts session: project evaluation, restore, and reading the restore output all see the same properties',
+    },
+    verbose: {
+      type: 'boolean',
+      description: 'Print debug messages',
+    },
+  },
+  help: (command, config) => `
+    Usage
+      $ ${command} [options] [CWD=.]
+
+    Options
+      ${getFlagListOutput(config.flags)}
+
+    Emits a single \`.socket.facts.json\` describing the resolved dependency
+    graph of the .NET solutions/projects at the top level of the given
+    directory. A bundled tool runs one MSBuild session — evaluate, restore,
+    read the restore output via NuGet's own APIs — so results reflect exactly
+    what NuGet resolved. A restore failure is a fatal error; pass
+    --ignore-unresolved to warn instead.
+
+    Each target framework a project restores is resolved separately: pass
+    --target-frameworks / --exclude-target-frameworks (comma-separated glob
+    patterns) to control which target frameworks are included (e.g.
+    --target-frameworks='net8.0'). RID-specific targets like net8.0/win-x64
+    match under their base target framework.
+
+    --dotnet-opts takes MSBuild property tokens (\`-p:Key=Value\`), applied to
+    the WHOLE session so evaluation, restore, and the emitted graph can never
+    disagree. Restore-specific settings have property forms, e.g.
+    \`-p:RestoreSources=<url>\` or \`-p:RestoreConfigFile=<path>\`.
+
+    Requires a .NET SDK (6.0 or newer). Legacy \`packages.config\` projects
+    are supported from the manifest itself (it pins the full closure): the
+    graph is flat — every package is listed as a direct dependency — and
+    \`developmentDependency="true"\` packages are marked dev. No restore is
+    attempted for them.
+
+    Support is beta. Please report issues or give us feedback on what's missing.
+
+    Examples
+
+      $ ${command} .
+      $ ${command} --target-frameworks='net8.0' .
+      $ ${command} --dotnet-opts='-p:Configuration=Release' .
+  `,
+}
+
+export const cmdManifestDotnet = {
+  description: config.description,
+  hidden: config.hidden,
+  run,
+}
+
+async function run(
+  argv: string[] | readonly string[],
+  importMeta: ImportMeta,
+  { parentName }: CliCommandContext,
+): Promise<void> {
+  const cli = meowOrExit({
+    argv,
+    config,
+    importMeta,
+    parentName,
+  })
+
+  const { json = false, markdown = false } = cli.flags
+
+  const dryRun = !!cli.flags['dryRun']
+
+  const outputKind = getOutputKind(json, markdown)
+
+  let [cwd = '.'] = cli.input
+  // Note: path.resolve vs .join:
+  // If given path is absolute then cwd should not affect it.
+  cwd = path.resolve(process.cwd(), cwd)
+
+  const sockJson = readOrDefaultSocketJson(cwd)
+
+  debugFn(
+    'inspect',
+    `override: ${SOCKET_JSON} dotnet`,
+    sockJson?.defaults?.manifest?.dotnet,
+  )
+
+  let {
+    bin,
+    dotnetOpts,
+    excludeTargetFrameworks,
+    ignoreUnresolved,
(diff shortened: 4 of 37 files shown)


jdalton:
@jfblaa Would this live better in the coana cli?

jdalton:
Six verified issues, most severe first. Each one is traced to the exact lines on this branch and checked against the v1.x base (and the pinned coana builds where that mattered). Two things that looked like bugs turned out fine — they're at the bottom so nobody re-chases them.


<details>
<summary><abbr title="Critical: fix before merge/run">🔴</abbr> <b>1. Windows line endings corrupt the last field of every record the C# tool emits</b></summary>
<blockquote>

`RecordsWriter.cs` writes records with `StreamWriter.WriteLine`. On Windows that ends each line with `\r\n`. The TypeScript parser in `records.mts` splits the stream on `'\n'` only and never strips the `'\r'`, so on a Windows machine every record's final field arrives with a stray `\r` character stuck to it.

The grammar puts meaningful data in that last position, so everything breaks quietly at once:

- Every `root` record's prod flag parses as `"1\r"`, which `bool()` reads as false — all roots get demoted to dev.
- Every `node`'s direct flag does the same — direct vs. transitive info is lost.
- Every `edge`'s child id becomes `"pkg:1.0\r"`, which matches no node key — the whole dependency graph flattens.
- Every `file`/`projectTgt` path fails the exists-check filter — the reachability sidecar loses all artifacts.

No error is raised anywhere; the scan "succeeds" with wrong data. The other three emitters already guard against this — the Gradle init script joins with `'\n'` explicitly and the Scala plugin appends `'\n'` — so the new C# writer is the one deviant. Windows is squarely in scope: the tool and CLI both carry Windows-specific handling elsewhere.

Fix idea 💡: set `_writer.NewLine = "\n"` in the `RecordsWriter` constructor. Worth also trimming a trailing `\r` per line in `records.mts` so the parser tolerates any future emitter that forgets — belt and suspenders on both sides of the protocol.

</blockquote>
</details>

<details>
<summary><abbr title="Significant: should be addressed">🟠</abbr> <b>2. The sidecar schema break is wider than the draft warning says — every reachability scan fails, not just polyglot ones</b></summary>
<blockquote>

The draft warning says a polyglot .NET+JVM `--reach` scan would break until the coana side ships. It's actually all of them. `sidecar.mts` now stamps `ecosystem` on every entry unconditionally — maven entries included. Both relevant pinned coana builds (15.6.7 on this branch, 15.8.4 on current v1.x) validate the sidecar with a schema that rejects any key it doesn't know, and neither knows `ecosystem`. The parse throws.

So a plain single-ecosystem Gradle or Maven `--reach` scan on this branch dies at the sidecar handoff, with no .NET anywhere in sight. The draft gate itself is right (don't merge until the coana release + pin bump), but anyone testing this branch should know the real blast radius.

Fix idea 💡: correct the PR body to say all `--reach` scans are gated — or only emit `ecosystem` once the pinned coana accepts it, which keeps pure-JVM reach working on this branch today.

</blockquote>
</details>

<details>
<summary><abbr title="Significant: should be addressed">🟠</abbr> <b>3. A tool crash after partial output still exits 0 — a truncated SBOM is reported as success</b></summary>
<blockquote>

The C# tool's catch-all prints to stderr and returns exit code 1, leaving whatever records were already flushed on disk — and writes no `failure` record. On the TypeScript side, the only check of that non-zero exit requires components AND projects AND failures AND unscannables to ALL be empty before treating the run as crashed.

Walk through what that means: the tool crashes while reading project 3 of 5. Projects 1–2 are already in the records file, so the all-empty guard passes, the CLI writes `.socket.facts.json`, logs success, and exits 0 — with projects 3–5 silently missing. The tool's own comment promises "the CLI treats it as a crashed build", but that stops being true the moment any partial output exists.

Fix idea 💡: emit a `failure` record from the catch-all (the record stream is still open there), or treat a non-zero tool exit as blocking on the TypeScript side no matter how much partial output landed.

</blockquote>
</details>

<details>
<summary><abbr title="Significant: should be addressed">🟠</abbr> <b>4. A failed restore can report success with stale data</b></summary>
<blockquote>

After the restore build, errors that belong to restore-unsupported projects are deliberately skipped as noise — but they stay in the error collection. The "restore failed" fallback record only fires when that collection is empty. Put those together: a restore that failed with only noise-classified errors emits no failure record at all.

The stale-assets backstop then turns quiet into wrong. If an earlier successful restore left `obj/project.assets.json` behind, the file-exists check passes and the tool reports that stale dependency closure as a fresh, successful scan — exit 0, zero failure records. The trigger is the exact scenario the noise filter's own comment cites: a solution with a legacy web project whose VS-only imports fail evaluation on Linux, in a checkout that restored successfully once before.

Fix idea 💡: gate the fallback on "no failure records were emitted" rather than the raw captured-error count — count what you reported, not what you caught.

</blockquote>
</details>

<details>
<summary><abbr title="Significant: should be addressed">🟠</abbr> <b>5. A leftover packages.config silently wins over the real PackageReference graph</b></summary>
<blockquote>

`IsRestoreSupported` checks for a `packages.config` file FIRST and returns false before ever looking at the project's `PackageReference` items; `ReadProject` routes the same way. A project that migrated to PackageReference but still has a leftover `packages.config` in its directory — a common migration remnant — gets scanned via the flat legacy path. The stale pinned list from the leftover file becomes the reported closure (all direct, no edges) while the project's actual dependency graph is never emitted. No failure record, exit 0.

NuGet itself resolves this the other way around: its restore-style logic checks PackageReference items before the packages.config file, so real `dotnet restore` on such a project restores the PackageReference graph. The tool's comment says it mirrors NuGet's eligibility model; the order is inverted.

Fix idea 💡: match NuGet's precedence (PackageReference first, packages.config as the fallback), and emit a warning or failure record when a project has both — that state is worth surfacing no matter which side wins.

</blockquote>
</details>

<details>
<summary><abbr title="Significant: should be addressed">🟠</abbr> <b>6. Polyglot repos overwrite one ecosystem's facts with another's</b></summary>
<blockquote>

Every facts leg writes the same `<cwd>/.socket.facts.json` with a plain overwrite, and the auto-manifest order is sbt → gradle → dotnet → maven. A repo with both a top-level `.csproj` and a `pom.xml` runs the dotnet leg, then the maven leg lands last and wipes the dotnet facts. Same story in a gradle+dotnet repo, where dotnet wipes gradle. The reachability sidecar accumulates across legs; the facts SBOM has no equivalent merge, and the scan flow reads the file once after all legs finish.

This collision already exists on v1.x — a gradle+maven repo clobbers today — so this PR extends it to a fourth leg rather than introducing it. But mixed .NET+JVM repos are this PR's headline scenario, so the pre-existing gap goes live for exactly the users the feature targets.

Fix idea 💡: accumulate facts across legs the way the sidecar already does — merge components/projects per leg into one document, written once at the end. That fixes the pre-existing gradle+maven case in the same motion.

</blockquote>
</details>

<details>
<summary><abbr title="Moderate/minor: worth addressing">🟡</abbr> <b>Smaller items</b></summary>
<blockquote>

- <abbr title="Moderate/minor: worth addressing">🟡</abbr> The TSV record grammar is hand-maintained in four languages (the comment in `records.mts`, plus the Groovy, Scala/Java, and C# emitters) with positional empty-string placeholders and no shared definition or cross-language golden test. Item 1 _(Windows line endings)_ is exactly the drift this invites — one emitter deviating from the protocol with nothing to catch it. A shared vector file all four emitters' tests consume would pin the contract.
- <abbr title="Moderate/minor: worth addressing">🟡</abbr> A stray top-level `.csproj` on a machine without the dotnet SDK now aborts the whole auto-manifest run (spawn ENOENT → abort). The JVM legs already behave this way when their tools are missing, and the detection comment shows this was a considered trade — but a cheap SDK-presence precheck with a clean "skipping dotnet: no SDK found" (or a pointer at `defaults.manifest.dotnet.disabled`) would spare the newly-detected repos a hard failure.
- <abbr title="Moderate/minor: worth addressing">🟡</abbr> packages.config artifact downloads run one at a time and buffer each whole `.nupkg` into a `MemoryStream`. Large packages (CUDA/ML runtimes ship multi-hundred-MB nupkgs) can OOM the tool on constrained runners — landing in the same catch-all path as item 3 _(crash exits 0)_. Bounded-concurrency downloads spooled to temp files bound both time and memory.
- <abbr title="Moderate/minor: worth addressing">🟡</abbr> MSBuild evaluation work repeats more than the "one MSBuild session" framing suggests: a fresh `ProjectCollection` per project in the read pass (which defeats shared SDK import caching), one extra inner-build evaluation per TargetFramework alias under `--with-files`, `MaxNodeCount` left at 1 so multiple top-level entries restore serially, and a timeout unwind that waits up to 30s per submission, one after another. Big multi-targeting solutions pay all of these at once.
- <abbr title="Moderate/minor: worth addressing">🟡</abbr> The success summary hand-codes `ecosystem === 'dotnet' ? 'target framework(s)' : 'configuration(s)'` while this same PR adds the per-ecosystem `configNoun` to the resolution-report dialect. The next ecosystem has to remember a buried ternary or its summary quietly says "configuration(s)". Read the noun from the dialect.
- <abbr title="Moderate/minor: worth addressing">🟡</abbr> The ".NET root marker" file-set (sln/slnx/csproj/fsproj/vbproj) is defined twice — `DOTNET_ROOT_FILE_RE` in TypeScript and `Discover()`/`IsProjectFile` in C# — with no test tying them together. Adding a project type to one side only desynchronizes detection from scanning.
- <abbr title="Moderate/minor: worth addressing">🟡</abbr> `purlTypeForTool` maps every non-dotnet tool to maven via a binary ternary. An exhaustive per-tool map makes the compiler force a decision for tool #5, instead of letting a future ecosystem silently assemble maven-typed components.
- <abbr title="Moderate/minor: worth addressing">🟡</abbr> `classifyNugetFailure` string-matches C# loader-failure signatures ("could not load file or assembly", an HRESULT) in TypeScript. When the C# side's failure wording shifts, misclassification sends users chasing feed problems. The tool already owns the failure record — let it self-classify with a category field.
- <abbr title="Moderate/minor: worth addressing">🟡</abbr> The config-glob header claims identical behavior across JS, Java, and .NET regex engines, but only a JS vitest suite exercises the vectors. Four consumers lean on that claim; running the same vector table under Java and .NET in the existing compat smokes would make it true by test rather than by assertion. One latent divergence exists already — .NET's `$` also matches before a trailing newline — though today's config/TFM names can't carry one, which is why this is a test ask and not a bug report.
- <abbr title="Moderate/minor: worth addressing">🟡</abbr> An uncompilable pattern used to degrade to a literal match (matches almost nothing); the shared contract now drops it, and since an empty include list means include-everything, a dropped include inverts into scanning all configurations. The CLI validates each source it emits, so this only bites out-of-band callers driving the shipped scripts directly — worth one line in the contract comment stating the direction change is intentional.
- <abbr title="Moderate/minor: worth addressing">🟡</abbr> `populateFilesFor` is forwarded to all three JVM tools but not to the dotnet tool (no flag in `ToolOptions.cs`). Nothing in the CLI sets it today, so this is a latent contract gap rather than a live bug — worth closing when the coana-side integration lands, since coana's own invocations are the intended caller.
- <abbr title="Moderate/minor: worth addressing">🟡</abbr> `EmitProjectTargets` uses one-arg `Path.GetFullPath(targetPath)` (anchored to the tool's working directory) where the bare-project and packages.config paths use the two-arg form anchored to the project directory. MSBuild guarantees an absolute `TargetPath` on the affected path, so this is a consistency fix, not a bug — the two-arg form just removes the trap.
- <abbr title="Moderate/minor: worth addressing">🟡</abbr> Duplication that will drift: the flag → socket.json-default → hardcoded-default ladder is copy-pasted a fifth time in `cmd-manifest-dotnet.mts` (~75 lines); the setup wizard inlines a fourth copy of the prompt/assign/delete block that `setupFactsOptions` already parameterizes; `convertDotnetToFacts` is a fourth rename-only passthrough; and `assertDotnetToolBuilt` clones `assertMavenExtensionBuilt` with different error text. Each is small; together they make the sixth ecosystem a ~200-line copy-paste tax.

</blockquote>
</details>

<details>
<summary><abbr title="Verified fine / informational">🟢</abbr> <b>Verified fine: two things that looked like bugs but aren't</b></summary>
<blockquote>

- <abbr title="Verified fine / informational">🟢</abbr> Windows caret-mangling of the compiled regex patterns (`^(?:...)$` passed as `-P`/`-D` values) can't happen: the spawn chain never uses a shell, so arguments reach the JVM untouched, and `.bat`/`.cmd` resolution either fails loudly or is rejected by Node's EINVAL hardening. There is no silent regex-corruption path.
- <abbr title="Verified fine / informational">🟢</abbr> The hand-rolled glob compiler in `config-glob.mts` is not a duplication of micromatch: micromatch compiles `*` to `[^/]*?` (which breaks the composite `net8.0/win-x64` names the dotnet tool deliberately matches) and doesn't support `[!...]` negation at all. It computes different answers, not just different syntax. The bespoke portable-subset compiler is the right call.

</blockquote>
</details>

Items 1 _(Windows line endings)_ and 3–5 _(the fail-closed trio)_ are the ones to settle before un-drafting — item 1 _(Windows line endings)_ because it's a one-line fix for a whole platform, and items 3 _(crash exits 0)_, 4 _(stale-restore success)_ and 5 _(packages.config precedence)_ because they all break the same promise the tool's design leans on: a wrong scan never exits green. Item 2 _(sidecar scope)_ just needs the PR body corrected so branch testers know what to expect.

jdalton:
[agent] Closing in favor of SocketDev/socket-facts#5 — the dotnet emitter and the centralized config-name globs now land in socket-facts, which owns the build-tool emitters and the `.socket.facts.json` contract that socket-cli and Coana both consume. The six verified findings in the review above are traced to exact lines, so they carry over: worth checking each against the socket-facts branch before it merges (the Windows line-ending corruption and the glob-centralization notes in particular map directly).

jdalton:
This moved to the facts repo and landed there: [SocketDev/socket-facts#5](https://github.com/SocketDev/socket-facts/pull/5), squashed as `d7eddc3`. @jfblaa, your authorship is preserved on the commit — thank you, the single-MSBuild-session design and the packages.config handling came over intact.

What changed on the way: four fail-closed fixes from the review (LF line endings in the records writer, a failure record on crash, the restore fallback gated on failures reported rather than errors captured, and PackageReference taking precedence over a leftover packages.config), plus a bump of the tool's NuGet references from 6.0.0 to the patched 6.0.6 — Socket's own scanner blocked the PR on [GHSA-68w7-72jg-6qpp](https://github.com/advisories/GHSA-68w7-72jg-6qpp).

The CLI half — the `socket manifest dotnet` command, auto-detection, the setup wizard, and the `socket.json` defaults — stayed here and can re-land as a small PR pinned against a facts release. One constraint carried over unchanged: the sidecar's new `ecosystem` tag needs the reachability sidecar's nuget schema change released before any reachability scan works, and that gate is wider than the original note said — it affects every reach scan, not only polyglot ones.

PR #1410, opened 2026-07-10
ci: restore provenance.yml publish workflow on main

barslev:
## Problem

The **"📦 Publish to npm registry"** workflow (`.github/workflows/provenance.yml`) has disappeared from the Actions UI, so the `socket` 1.1.x line can no longer be published from the sidebar.

Root cause: commit `f6165b07e` ("chore(ci): land 3 ci changes", part of the v2 monorepo cascade landed on `main` on 2026-07-09) **removed `provenance.yml` from `main`**. Because `main` is the default branch, GitHub only registers/surfaces workflows that exist there — so removing it:

- dropped **"📦 Publish to npm registry"** from the Actions sidebar (no entry, no "Run workflow" button), and
- makes `prune-workflow-runs.yml` eligible to wipe all of its run history (it deletes all runs for workflows whose source is gone from the default branch).

That workflow is still the **only** publisher for the `socket` 1.1.x line — currently `socket@latest = 1.1.140` on npm, shipped from the `v1.x` branch. The two publish workflows now on `main` (`npm-publish.yml`, `publish-npm.yml`) belong to the v2 `socket-cli-monorepo` and do **not** publish the 1.1.x line.

## Change

Restore `.github/workflows/provenance.yml` on `main`, **verbatim from `v1.x`**. This re-registers the workflow so it reappears in the Actions UI with a working "Run workflow" button.

## How to use after merge

Dispatch **"📦 Publish to npm registry"** and select the **`v1.x`** branch as the ref (the run uses the workflow definition and code from the selected ref). The `main` copy exists only so GitHub registers/surfaces the workflow.

## Note — this is a stopgap

The removal came from the automated fleet/template cascade. Unless the cascade template is updated to keep this file (or the team decides on a different publishing path for the 1.1.x line while v2 isn't yet `latest`), a future cascade may drop it again. Flagging separately to @jdalton.

---

> [!NOTE]
> **Medium Risk**
> Restores release automation that uses NPM OIDC/token publishing and can create git tags; operational risk if dispatched against the wrong ref, but logic is an unchanged restore from `v1.x`.
> 
> **Overview**
> Re-adds **`.github/workflows/provenance.yml`** on `main` so GitHub registers **“Publish to npm registry”** again in the Actions UI (workflows on the default branch only). The file matches the `v1.x` copy; after merge, operators should dispatch it with the **`v1.x`** ref so builds and publishes still use the 1.1.x line.
> 
> The restored workflow is the manual **build → npm publish (with provenance)** path for the `socket` / legacy / Sentry dist variants, plus idempotent **`v<version>`** tagging when the primary `socket` publish succeeds. It does not change application code—only brings back the workflow definition that was removed from `main` during the v2 CI cascade.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f9385b7c3af373bcc7f436e2be29eb2ae5e562de. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/provenance.yml
+++ b/.github/workflows/provenance.yml
@@ -1,0 +1,309 @@
+name: Publish to npm registry
+
+on:
+  workflow_dispatch:
+    inputs:
+      dist-tag:
+        description: 'npm dist-tag (latest, next, beta, canary, backport, etc.)'
+        required: false
+        default: 'latest'
+        type: string
+      debug:
+        description: 'Enable debug output'
+        required: false
+        default: '0'
+        type: string
+        options:
+          - '0'
+          - '1'
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    name: Build and Publish
+    runs-on: ubuntu-latest
+
+    permissions:
+      # `contents: write` needed to create the v<version> tag via gh api
+      # at the end of this job. Token is scoped to the dedicated tag step
+      # via GH_TOKEN env; never persisted in `.git/config` (checkout keeps
+      # persist-credentials: false so build/install steps can't reach it).
+      contents: write
+      id-token: write # NPM trusted publishing via OIDC
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2 (2026-05-20)
+        with:
+          persist-credentials: false
+
+      - name: Install pnpm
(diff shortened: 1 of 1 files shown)


jdalton:
Ah yes, migrating to a npm-publish.yml workflow but until it is up solid lets keep the old.

PR #1425, opened 2026-07-24
feat(scan): serve `socket scan view` from cached results with 202 polling (SURF-742)

jdalton:
## What this does

`socket scan view` shows the results of a full scan. Today it asks the API to compute those results from scratch every time. This change makes it read a **pre-computed (cached) copy** instead, which is faster and cheaper.

It does that by adding `?cached=true` to the request. The server has two possible answers:

- **200 OK** — the cached results are ready. The body is the usual ndjson (one JSON object per line), and we parse and show it exactly like before.
- **202 Accepted** — the cached results aren't ready yet. The server has kicked off a background job to build them and returns `{"status":"processing"}`. When this happens the CLI waits a moment and asks again, backing off a little each time (1s, then 2s, … up to 10s, giving up after 10 minutes). The user never sees the "processing" state — they just wait a bit longer and get the finished scan.

**Which modes change:** the default view and `--json` (the non-streaming modes that already buffer the whole response). The live `--stream --json` path is intentionally left untouched.

## Why it's safe to ship on its own

This is self-contained. It uses the CLI's own HTTP helper (`queryApiSafeTextWithStatus`) and does **not** call the Socket SDK for this path, so it stays on the currently-pinned `@socketsecurity/sdk` (1.4.96) and needs **no SDK release**. CI is green on its own.

The separate effort to move socket-cli `v1.x` onto the latest SDK (4.x) is tracked as **SURF-1446** — that is not required by, and does not block, this PR.

<details>
<summary>Implementation detail & parity with <code>main</code></summary>

- New helper `queryApiSafeTextWithStatus` in `utils/api.mts` returns `{ status, text }` so the caller can tell a 202 apart from a 200 (the existing `queryApiSafeText` swallowed the status). `queryApiSafeText` now delegates to it, so its behavior is unchanged for every other caller.
- `fetchScan` builds the `?cached=true` request, runs the 202→200 backoff loop, and parses the 200 ndjson via an exported `parseArtifactsNdjson`.
- This mirrors how `main`'s CLI already does cached scan reads (commits `e5fb60b1` + `f6741d2d`), adapted to the older v1.x layout. It deliberately does **not** use the SDK's cached `getFullScan`, because that method returns scan *metadata*, not the artifact list `scan view` renders.
- `--stream --json` still calls the streaming full-scan method unchanged.

</details>

<details>
<summary>Verification</summary>

- `tsgo`, `eslint`, `oxlint`, `biome`: clean.
- `vitest src/commands/scan/fetch-scan.test.mts`: 5 tests via `nock` against the real HTTP path — 200 cache hit, 202→200 poll, 404 error, plus `parseArtifactsNdjson` valid/invalid line cases. No owned-module mocking.
- `vitest src/utils/api.test.mts`: 10 tests, no regression from the `queryApiSafeText` refactor.

</details>

Linear: SURF-742. Do not merge yet.

### DIFF
--- a/src/commands/scan/fetch-scan.mts
+++ b/src/commands/scan/fetch-scan.mts
@@ -1,47 +1,78 @@
 import { debugDir, debugFn } from '@socketsecurity/registry/lib/debug'
 
-import { queryApiSafeText } from '../../utils/api.mts'
+import { queryApiSafeTextWithStatus } from '../../utils/api.mts'
 
 import type { CResult } from '../../types.mts'
 import type { SocketArtifact } from '../../utils/alert/artifact.mts'
+
+// HTTP 202 Accepted: cached results are still being computed; poll again.
+const HTTP_STATUS_ACCEPTED = 202
+
+export const CACHED_POLL_INITIAL_DELAY_MS = 1000
+export const CACHED_POLL_MAX_DELAY_MS = 10_000
+export const CACHED_POLL_TIMEOUT_MS = 10 * 60 * 1000
+
+function sleep(ms: number): Promise<void> {
+  return new Promise(resolve => {
+    setTimeout(resolve, ms)
+  })
+}
 
 export async function fetchScan(
   orgSlug: string,
   scanId: string,
 ): Promise<CResult<SocketArtifact[]>> {
-  const result = await queryApiSafeText(
-    `orgs/${orgSlug}/full-scans/${encodeURIComponent(scanId)}`,
-    'a scan',
-  )
-
-  if (!result.ok) {
-    return result
+  // Serve pre-computed results from the immutable store (`?cached=true`): a
+  // 200 carries the ndjson body, a 202 means the server enqueued a background
+  // job to compute them — poll with backoff until the results are ready, so
+  // callers only ever observe the final scan.
+  const path = `orgs/${orgSlug}/full-scans/${encodeURIComponent(scanId)}?cached=true`
+  const deadline = Date.now() + CACHED_POLL_TIMEOUT_MS
+  let delayMs = CACHED_POLL_INITIAL_DELAY_MS
+  for (;;) {
+    // eslint-disable-next-line no-await-in-loop
+    const result = await queryApiSafeTextWithStatus(path, 'a scan')
+    if (!result.ok) {
+      return result
+    }
+    if (result.data.status !== HTTP_STATUS_ACCEPTED) {
+      return parseArtifactsNdjson(result.data.text)
+    }
+    if (Date.now() >= deadline) {
+      return {
+        ok: false,
+        message: 'Scan results not ready',
+        cause: `The Socket API is still computing cached results for scan ${scanId} after ${CACHED_POLL_TIMEOUT_MS / 60_000} minutes (path: ${path}). Retry in a few minutes — the server keeps computing in the background.`,
+      }
+    }
+    // eslint-disable-next-line no-await-in-loop
+    await sleep(delayMs)
+    delayMs = Math.min(delayMs * 2, CACHED_POLL_MAX_DELAY_MS)
   }
+}
 
-  const jsonsString = result.data
-
-  // This is nd-json; each line is a json object
+export function parseArtifactsNdjson(
+  jsonsString: string,
+): CResult<SocketArtifact[]> {
+  // This is nd-json; each line is a json object.
   const lines = jsonsString.split('\n').filter(Boolean)
-  let ok = true
-  const data = lines.map(line => {
+  const data: SocketArtifact[] = []
+
+  for (let i = 0, { length } = lines; i < length; i += 1) {
+    const line = lines[i]!
     try {
-      return JSON.parse(line)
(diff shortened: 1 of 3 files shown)


jdalton:
Withdrawing — the cached scan-view behavior (SURF-742) will land as part of bumping socket-cli v1.x to the latest @socketsecurity/sdk (4.x) under SURF-1446, rather than a bespoke direct-API poll. Superseding this branch.

src/commands/scan/fetch-scan.mts
```diff
@@ -1,47 +1,55 @@
...
-    return { ok: true, data }
   }
 
   return {
-    ok: false,
-    message: 'Invalid Socket API response',
-    cause:
-      'The Socket API responded with at least one line that was not valid JSON. Please report if this persists.',
+    ok: true,
+    data: result.data as unknown as SocketArtifact[],
```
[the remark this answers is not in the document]
jdalton: Moot on the current diff (c57dc34c5). `fetchScan` was reworked to the direct-API path via `queryApiSafeTextWithStatus`, which provides the shared spinner, try/catch, and `CResult` error mapping (the same helper the rest of the CLI uses for direct GETs) — and it no longer calls the SDK at all. So there is no `getOrgFullScanCached`/`handleApiCall` bypass left to address. This mirrors how `main`'s CLI does cached scan reads.

src/commands/scan/fetch-scan.mts
```diff
@@ -1,47 +1,55 @@
...
+    ]
+      .filter(Boolean)
+      .join(' ')
+    return {
+      ok: false,
+      message: 'Socket API error',
+      cause,
+      data: {
+        code: result.status,
+      },
```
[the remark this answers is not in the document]
jdalton: Moot on the current diff (c57dc34c5). `fetchScan` no longer calls an SDK method — it goes through `queryApiSafeTextWithStatus`, whose internal try/catch turns network and polling failures into a failed `CResult` rather than an unhandled throw. The 202 poll loop only continues on an ok result and returns the error `CResult` otherwise.

src/commands/scan/fetch-scan.mts
```diff
@@ -1,47 +1,78 @@
...
-
-  if (!result.ok) {
-    return result
+  // Serve pre-computed results from the immutable store (`?cached=true`): a
+  // 200 carries the ndjson body, a 202 means the server enqueued a background
+  // job to compute them — poll with backoff until the results are ready, so
+  // callers only ever observe the final scan.
+  const path = `orgs/${orgSlug}/full-scans/${encodeURIComponent(scanId)}?cached=true`
+  const deadline = Date.now() + CACHED_POLL_TIMEOUT_MS
+  let delayMs = CACHED_POLL_INITIAL_DELAY_MS
```
bcomnes, line 31: Should double check this against what the actual full scan timeout is in depscan.

src/commands/scan/fetch-scan.mts
```diff
@@ -1,47 +1,78 @@
...
-    'a scan',
-  )
-
-  if (!result.ok) {
-    return result
+  // Serve pre-computed results from the immutable store (`?cached=true`): a
+  // 200 carries the ndjson body, a 202 means the server enqueued a background
+  // job to compute them — poll with backoff until the results are ready, so
+  // callers only ever observe the final scan.
+  const path = `orgs/${orgSlug}/full-scans/${encodeURIComponent(scanId)}?cached=true`
```
bcomnes, line 29: Any reason this isn't implemented in the SDK and then just consumed in the CLI? Clients of that shouldn't need to implement this logic.

bcomnes (approved):
We should definitely validate this on real data before releasing, and also this seems preferable to live in the SDK, so all clients benefit from the improvement, but otherwise on target. Please break out additional tickets for the SDK if you think thats a reasonable direction to take and even consider landing it via that approach, but if you want to just get this out as an intermediate step that also seems fine.

src/commands/scan/fetch-scan.mts
jdalton: No reason, and that is where it ended up. This PR is withdrawn in favor of the SDK owning the behavior: `socket-sdk-js` now ships the cached-scan poll helper (`src/utils/poll.mts`), so every client gets the 202 handling for free, and the CLI consumes it through the sdk 4.x bump in #1427.

jdalton:
Closing as superseded by the sdk 4.x bump in #1427, which carries the SURF-742 cached scan-view behavior via the SDK's own poll helper.

src/commands/scan/fetch-scan.mts
jdalton: Checked server-side: there is no fixed completion deadline to align with — the background compute is retried with backoff and its runtime scales with scan size, so any fixed client-side deadline is a heuristic. The SDK path that supersedes this PR makes the deadline explicit and configurable: socket-sdk-js's `src/utils/poll.mts` drives the 202 loop with `maxPollMs` (default `DEFAULT_POLL_TIMEOUT` = 5 minutes, `src/constants.mts`).

PR #1457, opened 2026-08-01
fix(deps): bump the maven smoke fixture's commons-io past GHSA-78wr-2p64-hpwj

jdalton:
This clears the one open Dependabot alert on the repo. The Maven smoke-test fixture pinned `commons-io` **2.11.0**, which falls in the vulnerable range for [GHSA-78wr-2p64-hpwj](https://github.com/advisories/GHSA-78wr-2p64-hpwj) (a denial-of-service on untrusted input to `XmlStreamReader`, affecting `>= 2.0, < 2.14.0`). It is now **2.22.0**, the current release.

This is test-fixture code, not anything we ship, so there is no runtime exposure — but the alert is real and stays open until the pinned version moves.

While bumping it I noticed the version was written in **five** places: once in `app/pom.xml` and four times inside `smoke-test.sh` (the assertion in each of its two Python blocks, and the GAV scope file it writes). A future bump had four chances to go half-done and leave the assertions looking for a version the build no longer produces. The script now reads it from a single `COMMONS_IO_VERSION` variable and passes it into both Python blocks as an argument, so a bump is one edit in the pom and one in the script.

<details>
<summary>Validation</summary>

- `bash -n smoke-test.sh` — clean.
- Built the fixture end to end with the new version, using a throwaway local repository so the result did not depend on anything already cached: `mvn --batch-mode -Dmaven.repo.local=$(mktemp -d) compile` in `.../maven-compat/project` exits 0, and `commons-io-2.22.0.jar` resolves from Central.
- The rewritten assertion blocks were exercised against synthetic `records.tsv` input: they pass on well-formed records, and still fail correctly when commons-io is absent and when a test dependency wrongly lands in a prod root.
- **Correction to an earlier version of this description:** it said `smoke-test.sh` could not be run here because the Coana extension jar is "built elsewhere". That was wrong — `scripts/maven-extension/build-jar.sh` builds it from this repo. The follow-up PR stacked on this branch runs the full smoke test end to end against that jar, and both phases pass.

One thing worth knowing if you go looking: after a build, the local repository still contains `commons-io` 2.11.0. That is not this fixture. `maven-resources-plugin` and `maven-site-plugin` declare 2.11.0 among their own dependencies, so Maven fetches it to run the build itself. Dependabot alerts on the manifest's declared dependency, which is what this PR changes.

</details>

---

> [!NOTE]
> **Low Risk**
> Test fixture dependency and script assertions only; no production code or security-sensitive runtime paths.
> 
> **Overview**
> Bumps the Maven compatibility smoke fixture’s `commons-io` dependency from **2.11.0** to **2.22.0**, clearing the Dependabot alert for GHSA-78wr-2p64-hpwj. This is test-only fixture code, not shipped runtime.
> 
> The smoke script previously hard-coded that version in several places (Python checks and the `populateFilesFor` GAV file). It now uses a single **`COMMONS_IO_VERSION`** variable (must stay aligned with `app/pom.xml`) and passes it into both Python assertion blocks so future bumps need only two edits.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 71cadc3c9e3786273b6d0506fd71c8f2e14e36ac. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### Relationship to the other PRs

**Supersedes #1449** (Dependabot's `2.11.0 → 2.14.0`). That PR edits only the pom, which would leave `smoke-test.sh`'s four hardcoded copies of the old version asserting against a jar the build no longer produces. This PR goes to the current release instead of the minimum patched version, and removes the duplication that made the trap possible.

**Paired with #1458**, which ports the same fix to `v1.x`. That branch has its own copy of the fixture with the identical pin. Dependabot only scans the default branch, so the `v1.x` copy never appeared in the security tab despite the branch being actively developed.

### DIFF
--- a/packages/cli/src/commands/manifest/scripts/test/maven-compat/project/app/pom.xml
+++ b/packages/cli/src/commands/manifest/scripts/test/maven-compat/project/app/pom.xml
@@ -18,7 +18,7 @@
     <dependency>
       <groupId>commons-io</groupId>
       <artifactId>commons-io</artifactId>
-      <version>2.11.0</version>
+      <version>2.22.0</version>
     </dependency>
     <!-- test dep -> dev flag; pulls hamcrest transitively -->
     <dependency>
--- a/packages/cli/src/commands/manifest/scripts/test/maven-compat/smoke-test.sh
+++ b/packages/cli/src/commands/manifest/scripts/test/maven-compat/smoke-test.sh
@@ -9,6 +9,9 @@
 #  - materializes resolved external jars under -Dsocket.withFiles;
 #  - scopes that materialization to -Dsocket.populateFilesFor (a newline-delimited GAV file).
 #
+# Assertions match on groupId:artifactId and ignore the version, so bumping a fixture pom (a CVE, say)
+# needs no edit here. The poms are the only place a version is written.
+#
 # Usage: smoke-test.sh <path-to-mvn> <path-to-extension-jar>
 set -euo pipefail
 HERE="$(cd "$(dirname "$0")" && pwd)"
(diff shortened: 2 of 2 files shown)


jdalton:
Closing — this change is already on the target branch. The fixture now pins `commons-io` 2.22.0 and the smoke test carries no hardcoded version, which is exactly what this PR did; a diff of the touched paths against the base is empty.

The follow-up that replaces those external artifacts with generated stubs is still open and has been rebased so it no longer depends on this branch.

PR #1458, opened 2026-08-01
fix(deps): bump the maven smoke fixture's commons-io past GHSA-78wr-2p64-hpwj (v1.x)

jdalton:
Port of #1457 to `v1.x`, which carries the same vulnerable pin.

`v1.x` has its own copy of the Maven smoke fixture at `src/commands/manifest/scripts/test/maven-compat/`, and its `smoke-test.sh` is byte-identical to main's. Both pinned `commons-io` **2.11.0**, which is in the vulnerable range for [GHSA-78wr-2p64-hpwj](https://github.com/advisories/GHSA-78wr-2p64-hpwj) (`>= 2.0, < 2.14.0`). Now **2.22.0**.

Worth saying why this is a separate PR rather than something to skip: Dependabot only scans the default branch, so its alert names main's path and this copy never showed up in the security tab. But `v1.x` is actively committed to, so the same pin was sitting there unflagged.

As on main, the version is no longer written in the test at all. The assertions match on `groupId:artifactId` and ignore the version, and the scoped-materialization run derives its GAV from the records the first run emitted, so the poms are the only place a version appears.

<details>
<summary>Validation</summary>

- `bash -n smoke-test.sh` clean; no `2.11.0` remains anywhere in the fixture.
- Built the fixture end to end with a throwaway local repository (`mvn --batch-mode -Dmaven.repo.local=$(mktemp -d) compile`): exits 0, resolves `commons-io-2.22.0.jar` from Central.
- The rewritten assertion blocks were exercised against synthetic `records.tsv` input: they pass on well-formed records, and still fail correctly when commons-io is absent ("missing external prod dep commons-io") and when a test dependency wrongly lands in a prod root.
- **Correction to an earlier version of this description:** it said `smoke-test.sh` could not be run here because the Coana extension jar is "built elsewhere". That was wrong — `scripts/maven-extension/build-jar.sh` builds it from this repo. The follow-up PR stacked on this branch runs the full smoke test end to end against that jar, and both phases pass.

</details>

---

> [!NOTE]
> **Low Risk**
> Changes only affect the Maven smoke fixture and its shell assertions; no production manifest or extension logic is modified.
> 
> **Overview**
> Bumps **commons-io** in the Maven compat smoke fixture `app/pom.xml` from **2.11.0** to **2.22.0**, clearing [GHSA-78wr-2p64-hpwj](https://github.com/advisories/GHSA-78wr-2p64-hpwj) for the vulnerable `>= 2.0, < 2.14.0` range on the `v1.x` copy of the fixture.
> 
> **`smoke-test.sh`** no longer pins dependency versions in assertions. A `coord(prefix)` helper resolves emitted coordinate IDs by `groupId:artifactId` prefix, and the scoped `populateFilesFor` run builds its GAV list from the first run’s `records.tsv` instead of a hardcoded `commons-io:commons-io:2.11.0`. Future pom-only version bumps (e.g. CVE fixes) should not require editing the shell test.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3dcaae09edf8fcf5872ad979ce6e7104a9531309. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/commands/manifest/scripts/test/maven-compat/project/app/pom.xml
+++ b/src/commands/manifest/scripts/test/maven-compat/project/app/pom.xml
@@ -18,7 +18,7 @@
     <dependency>
       <groupId>commons-io</groupId>
       <artifactId>commons-io</artifactId>
-      <version>2.11.0</version>
+      <version>2.22.0</version>
     </dependency>
     <!-- test dep -> dev flag; pulls hamcrest transitively -->
     <dependency>
--- a/src/commands/manifest/scripts/test/maven-compat/smoke-test.sh
+++ b/src/commands/manifest/scripts/test/maven-compat/smoke-test.sh
@@ -9,6 +9,9 @@
 #  - materializes resolved external jars under -Dsocket.withFiles;
 #  - scopes that materialization to -Dsocket.populateFilesFor (a newline-delimited GAV file).
 #
+# Assertions match on groupId:artifactId and ignore the version, so bumping a fixture pom (a CVE, say)
+# needs no edit here. The poms are the only place a version is written.
+#
 # Usage: smoke-test.sh <path-to-mvn> <path-to-extension-jar>
 set -euo pipefail
 HERE="$(cd "$(dirname "$0")" && pwd)"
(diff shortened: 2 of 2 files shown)


jdalton:
Closing — this change is already on the target branch. The fixture now pins `commons-io` 2.22.0 and the smoke test carries no hardcoded version, which is exactly what this PR did; a diff of the touched paths against the base is empty.

The follow-up that replaces those external artifacts with generated stubs is still open and has been rebased so it no longer depends on this branch.

PR #1460, opened 2026-08-01
test(manifest): resolve the JVM compat fixtures' deps from generated stub artifacts

jdalton:
The JVM compat test fixtures pull three real, published artifacts down from Maven Central purely so the dependency graph has the right *shape* — one external production dependency, one test-scope dependency, and one transitive of that test dependency. No fixture code ever imports them. They are never compiled against and never executed. They are stage props.

Stage props still get CVEs. One of them, `commons-io:commons-io:2.11.0`, fell into the range of GHSA-78wr-2p64-hpwj, Dependabot opened #1449, and #1457 / #1458 are open right now purely to move a version number in a test fixture. That will happen again with the next advisory against whatever the fixtures happen to pin. This PR replaces those three artifacts with stubs generated at test time, so no advisory can ever touch them again.

The second thing it fixes: running these suites used to write into your own `~/.m2`, `~/.gradle` and `~/.ivy2`. That is untidy, but the real cost is that a run could pass because an artifact happened to already be warm in your home cache — so a genuine resolution regression could hide on your machine and only appear in CI. Each tool is now pointed at one shared cache root instead.

This is the `v1.x` half of #1459, mirroring how #1457 and #1458 were split. It builds on #1458 (`fix/commons-io-cve-v1x`), so the version-agnostic assertions that PR introduced are inherited rather than fought with. **Merge that one first.**

<details>

<summary><b>The problem</b> — three real artifacts pulled in for their shape, not their code</summary>

The Maven smoke fixture declared `commons-io:commons-io`, `junit:junit`, and junit's transitive `org.hamcrest:hamcrest-core`. It needed them only for their shape: one external prod dependency, one test-scope dependency, and one transitive of that test dependency.

No fixture code imports any of them. `App.java` calls `Lib.greet()` and nothing else, so they are never compiled against and never executed.

They still cost real work. commons-io 2.11.0 fell into the range of GHSA-78wr-2p64-hpwj, Dependabot opened #1449, and #1457 / #1458 exist just to move a number.

</details>

<details>

<summary><b>The fix</b> — a pattern <code>gradle-compat</code> has always used, moved up one directory so all three fixtures share it</summary>

`gradle-compat` has never had this problem. Its `make-localrepo.sh` generates a tiny file-based Maven repository of stub artifacts — a generated `.pom` plus an empty-but-valid `.jar` — at test time, and the Gradle fixture resolves `demo.lib:foo` and `demo.test:bar` from it. Nothing binary is committed and no advisory can ever touch it. `maven-compat` was the odd one out, not the place to invent something new.

So the generator moves up one directory to `test/make-stub-repo.sh`, next to `run-compat.sh`, and all three fixtures share it. It takes an output directory and a list of coordinates. The one new capability is that a stub can declare a dependency on another stub, which is what Maven needs for the transitive shape:

```sh
bash "$HERE/../make-stub-repo.sh" "$STUBS" \
  'demo.ext:tool:1.0' \
  'demo.ext:harness:1.0+demo.ext:harness-core:1.0' \
  'demo.ext:harness-core:1.0'
```

`maven-compat` now declares these instead:

| Stub coordinate | Role it plays in the fixture | Was |
| --- | --- | --- |
| `demo.ext:tool` | external prod dependency | `commons-io:commons-io` |
| `demo.ext:harness` | test-scope dependency | `junit:junit` |
| `demo.ext:harness-core` | transitive of the test-scope dependency | `org.hamcrest:hamcrest-core` |

`sbt-compat` had the same exposure — it still pinned the vulnerable `commons-io:commons-io:2.11.0`, invisible to Dependabot only because Dependabot does not read `build.sbt` — so it moves to the same stubs.

</details>

<details>

<summary><b>How resolution is wired</b> — one repository property on the root pom, overridden by the script</summary>

The root pom declares one repository whose URL comes from a property, and `smoke-test.sh` passes the absolute path of the repo it just generated:

```xml
<repository>
  <id>socket-stubs</id>
  <url>${stub.repo.url}</url>
  ...
</repository>
```

Child modules inherit it, which is verified below. The property has a default of `file://${maven.multiModuleProjectDirectory}/localrepo` so a bare `mvn compile` inside `project/` also works; `-Dstub.repo.url` from the script overrides it.

**Alternative considered and rejected:** seeding the artifacts straight into the run's local repository. That has to know the local repo path, which breaks the moment anyone passes `-Dmaven.repo.local` — including the empty-local-repo check below.

The generated repos are gitignored and rebuilt from scratch on every run, like `records.tsv`.

</details>

<details>

<summary><b>The caches</b> — one shared root, stable so the tool closure stays warm, stubs evicted every run</summary>

The suites no longer touch your own caches. `compat-cache.sh` resolves a single cache root and each tool is pointed at it:

```
${SOCKET_COMPAT_CACHE:-${TMPDIR:-/tmp}/socket-manifest-compat}
```

| Tool | How it is redirected | State before this PR |
| --- | --- | --- |
| Maven | `-Dmaven.repo.local` | wrote to `~/.m2` |
| Gradle | `-g` | already isolated, but at a directory inside the repo tree |
| sbt | `-Dsbt.ivy.home` and `COURSIER_CACHE` | global base was already a `mktemp -d`; the Ivy home and coursier cache were not covered |

`SOCKET_COMPAT_CACHE` already existed in `run-compat.sh` for the downloaded build-tool distributions, so this reuses it rather than inventing a second knob; `run-compat.sh` now reads it from the shared file too.

The most important reason is reproducibility, not tidiness — a run used to be able to pass off a warm home cache. Two lesser reasons: `records.tsv` embeds the resolved artifact path, so it used to contain a path under the developer's home directory, and a test should not write into a developer's real Maven cache.

**Trade-off: a stable root, not a fresh one per run.** A `mktemp -d` every time would be maximally hermetic, but it would re-download Maven's whole plugin closure on every invocation, making the suite slower and more network-dependent than it is today. So the root is stable and each tool's own closure is cached across runs — while the *stub* artifacts are evicted from it before every run, so the thing actually under test is always resolved from scratch. That keeps the fresh-resolution property exactly where it matters. Set `SOCKET_COMPAT_CACHE` to a fresh `mktemp -d` when you want a completely cold run.

The first run after the cache root is cleared pays for the tool's closure download. Measured here: about 70 seconds cold, about 35 warm.

</details>

<details>

<summary><b>Coverage is not weakened</b> — the same assertions, and one that gets stronger</summary>

The assertions prove exactly what they proved before: external prod dep in a prod root with a materialized jar, test-scope dep and its transitive kept out of prod roots, internal reactor module `demo:lib` emitted by its bare id and marked direct, and `-Dsocket.populateFilesFor` materializing only the in-scope artifact. Matching is still on `groupId:artifactId` with the version ignored, and the scoped run still derives its GAV from the emitted records, so no version is written anywhere in the test.

One assertion gets *stronger*. The transitive check used to be conditional (`if hamcrest and in_prod(hamcrest)`), tolerating hamcrest being absent because that depended on which junit version resolved. With a stub graph the transitive is deterministic, so its absence is now a failure.

</details>

<details>

<summary><b>What this does not claim</b> — the fixture is still not offline, and commons-io still lands in the local repo</summary>

This is not a fully offline fixture, and the change does not make it one. Maven fetches its own plugins from Central to run at all, and those plugins have their own dependencies. `maven-resources-plugin:3.3.1` depends on `commons-io:commons-io:2.11.0` — the very version this whole thread is about — so commons-io still lands in the local repository during a run.

It is not in our dependency graph, we do not declare it, and Dependabot does not alert on it. What this PR removes is *our* declared externals, not Maven's build closure.

</details>

<details>

<summary><b>What I checked</b> — Maven 3.9.10 and Gradle 8.14 end to end; sbt is reasoned, not run</summary>

Maven 3.9.10, Gradle 8.14, JDK 21.0.3 on macOS.

**This branch was exercised on its own, not just via #1459.** Both PRs carry the identical fixture change; the `v1.x` tree keeps these files at `src/commands/manifest/scripts/test/` and uses per-fixture `.gitignore` files instead of the root one, which is the only difference. The Maven and Gradle smoke tests were re-run end to end against this branch's own copy.

**Ran — the Maven smoke test end to end.** The brief assumed it could not run, because it needs the Coana extension jar as its second argument — but `maven-extension/build-jar.sh` builds that jar from this repo, so it was buildable locally. Both phases pass:

```
+ Apache Maven 3.9.10
PASS: tool=maven; demo.ext:tool:jar:1.0 prod+jar; harness/harness-core dev; internal demo:lib (bare id, direct)
PASS (populateFilesFor scoping): demo.ext:tool materialized, demo.ext:harness skipped
```

**Ran — the same thing again with an empty local repository** (`MAVEN_ARGS=-Dmaven.repo.local=$(mktemp -d)`), still exit 0. Counting where every artifact in that repository came from afterwards:

- 6 files from `socket-stubs` — the three stubs' poms and jars, including both test-scope jars materialized under `--with-files`
- 71 files from `central` — Maven's own plugin closure
- `commons-io` present at exactly one version, 2.11.0, `_remote.repositories` says `central=`, pulled by `maven-resources-plugin-3.3.1.pom`
- `org/junit/junit-bom` poms present, also `central=`, also plugin closure
- no `junit:junit` and no `hamcrest` anywhere

`mvn --batch-mode -Dmaven.repo.local=$(mktemp -d) dependency:tree` confirms the shape and that the child module inherited the repository:

```
demo:app:jar:1.0
+- demo:lib:jar:1.0:compile
+- demo.ext:tool:jar:1.0:compile
\- demo.ext:harness:jar:1.0:test
   \- demo.ext:harness-core:jar:1.0:test
```

**Ran — cache isolation, verified rather than assumed.** With the cache root pointed at a scratch path, the Maven smoke test passes both phases and the Gradle smoke test passes both phases. Full recursive listings of the real `~/.m2` and `~/.gradle` hash identically before and after those runs. Neither suite's `records.tsv` contains any path under the home directory; the Maven file record now points into the isolated repository and the Gradle one points straight at the stub repo.

**Ran — an A/B on the stub eviction.** Leaving the previous run's stub copy in the local repository and then deleting the generated stub repo, `mvn compile` still succeeds — that is precisely the masking this guards against. Doing what the script does, evicting the stubs first, the same run fails with `Could not find artifact demo.ext:tool:jar:1.0`. So a broken stub repo cannot be hidden by a warm cache.

The assertion blocks needed no changes for any of this; they key on the `.jar` suffix, not on any path.

**Ran — Gradle is not regressed.** `gradle-compat/smoke-test.sh` runs end to end on Gradle 8.14 and passes both phases. The repo the shared generator produces for the Gradle coordinates is byte-identical to the old script's: same paths, same pom bytes, jars differing only in zip timestamps.

**Ran — the assertion blocks against synthetic `records.tsv` input**, extracted from the committed script rather than a copy. 16 cases, all behaving as expected: well-formed records pass; the prod stub missing, its jar not materialized, the test-scope dep in a prod root, the transitive in a prod root, the transitive missing, `demo:lib` absent or not direct or dev-only, two versions of the same coordinate, and both scoped-run defects all fail with the right message. The sbt block got the same treatment on 4 cases.

**Ran —** `bash -n` and `shellcheck`, clean on all four scripts.

**Could not verify — the older matrix rows.** The compat matrix covers Maven 3.6.3 / 3.8.8 / 3.9.9 and Gradle 1.12 through 9.2.1; only Maven 3.9.10 and Gradle 8.14 are installed here, so the older rows are unverified.

**Did not run — sbt at all.** sbt is not installed here and there is no CI for this matrix, so the sbt fixture changes are reasoned, not run — both the stub coordinates and the `-Dsbt.ivy.home` / `COURSIER_CACHE` relocation. The reasoning: `"Socket Stubs" at s"file://.../localrepo"` is a `MavenRepository`, which coursier supports natively on sbt 1.x and Ivy resolves through `IBiblioResolver` on 0.13. That form was chosen over `Resolver.file(...)(Resolver.mavenStylePatterns)` precisely because the pattern-based form is the one with known coursier sharp edges. If you would rather not carry an unrun change, the sbt hunk is easy to split out.

</details>


jdalton:
[agent] Rebased onto v1.x and brought this in line with the main-line PR (#1459, now merged as abd61a577), plus two things specific to this branch.

Dropped 256 committed Gradle cache files. This branch had `gradle-compat/.gradle-home/` checked in — about 1.7 MB of `fileHashes.bin`, `*.lock` and compiled `_BuildScript_.class` files — because v1.x's `.gitignore` has no entries for the compat fixtures at all. The diff is 19 files now instead of 275.

So I added the missing ignore entries for every path the fixtures generate. One difference from main worth calling out: `.gradle-home/` stays ignored here. On main that line was dropped because the Gradle user home moved under the temp dir, but on this line `gradle-compat/smoke-test-workspaces.sh` line 14 still uses `GUH="$HERE/.gradle-home"`, and that script does not exist on main and is not touched by this PR. Dropping the line here would have let the same cache reappear as untracked files.

Also ported the internal-module assertion fix Bugbot caught on #1459: `coord()` takes a `fields` count and the bare-id lookup passes `fields=3`, so `demo:lib:` no longer matches `demo:lib:jar:1.0`. That block moved into `maven-compat/assert-records.py` with a unit test at `test/maven-compat-smoke-assertions.test.mts`.

jdalton:
Landed on v1.x as 6593c54c7 via the landing queue.

PR #1467, opened 2026-08-03
Add socket manifest dynamic-sbom-inference

jfblaa:
## Summary
- Adds `socket manifest dynamic-sbom-inference`: recursively discovers gradle/sbt/maven build roots and generates a facts SBOM per independent root, with reactor/multi-module coverage tracking so submodules aren't re-invoked.
- Adds a companion `socket manifest setup --dynamic-sbom-inference` wizard to scaffold per-project `socket.json` config across a tree, with a cascading `defaults.manifest.*` inheritance model and an explicit `null` sentinel for clearing an inherited value.
- Adds lightweight per-ecosystem workspace enumeration (no dependency resolution) used by the setup wizard's discovery step.

## Test plan
- [x] `pnpm check:tsc` / `pnpm check:lint`
- [x] `pnpm build:dist:src`
- [x] Relevant unit test suites pass

---

> [!NOTE]
> **Medium Risk**
> Large new surface around spawning Gradle/Maven/sbt and monorepo classification; mis-skips or fail-closed aborts could block SBOM generation on complex trees, but changes are localized to manifest CLI and bundled build-tool scripts.
> 
> **Overview**
> Adds **`socket manifest dynamic-sbom-inference`**, which walks a tree under CWD, finds Gradle/Maven/sbt build roots, and writes **`.socket.facts.json`** per independent root. Reactor members already covered by a parent’s facts **`subprojectDir`** list are skipped; a failed root aborts the walk so later roots aren’t misclassified.
> 
> **`socket manifest setup`** gains hidden **`--dynamic-sbom-inference`** (and **`--exclude-paths`**) to scan the tree, bulk-disable excluded paths, and interactively configure per-root **`socket.json`** with cascade inheritance and explicit **`null`** to clear inherited values.
> 
> Lightweight **workspace enumeration** (Gradle/sbt/Maven init scripts/plugins/extension) lists projects without resolving dependencies, used by the setup wizard’s coverage pruning. Dist build copies **`socket-workspaces`** assets alongside facts scripts.
> 
> Manifest commands now use a shared **`exclude-paths`** flag description, **`javaHome`** from **`socket.json`** (with **`$VAR`/`${VAR}`** expansion into **`JAVA_HOME`**), and **`runManifestFacts`** returns generated projects for recursive coverage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 36f354f08cd52930c0700b04030f0b963accfae2. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/commands/manifest/cmd-manifest-setup.mts
+++ b/src/commands/manifest/cmd-manifest-setup.mts
@@ -5,6 +5,7 @@ import { logger } from '@socketsecurity/registry/lib/logger'
 import { handleManifestSetup } from './handle-manifest-setup.mts'
 import constants, { SOCKET_JSON } from '../../constants.mts'
 import { commonFlags } from '../../flags.mts'
+import { cmdFlagValueToArray } from '../../utils/cmd.mts'
 import { meowOrExit } from '../../utils/meow-with-subcommands.mts'
 import { getFlagListOutput } from '../../utils/output-formatting.mts'
 
@@ -20,9 +21,23 @@ const config: CliCommandConfig = {
   hidden: false,
   flags: {
     ...commonFlags,
+    // Only meaningful alongside the hidden --dynamic-sbom-inference below; kept hidden too.
+    excludePaths: {
+      type: 'string',
+      isMultiple: true,
+      hidden: true,
+      description:
+        'Build roots matching these glob patterns (and everything beneath them) are marked disabled. Patterns are anchored micromatch globs matched relative to CWD: `legacy` matches only `<cwd>/legacy`; use `**/legacy` to match at any depth. Negation patterns (`!path`) are not supported. Accepts a comma-separated value or multiple flags.',
+    },
     defaultOnReadError: {
       type: 'boolean',
       description: `If reading the ${SOCKET_JSON} fails, just use a default config? Warning: This might override the existing json file!`,
+    },
+    dynamicSbomInference: {
+      type: 'boolean',
+      hidden: true,
+      description:
+        'After configuring CWD, recursively discover every gradle/sbt/maven build root beneath it and mark `disabled: true` on whatever matches --exclude-paths; everything else is left untouched',
     },
   },
   help: (command, config) => `
@@ -75,7 +90,7 @@ async function run(
     parentName,
   })
 
-  const { defaultOnReadError = false } = cli.flags
+  const { defaultOnReadError = false, dynamicSbomInference = false } = cli.flags
 
   const dryRun = !!cli.flags['dryRun']
 
@@ -89,5 +104,12 @@ async function run(
     return
   }
 
-  await handleManifestSetup(cwd, Boolean(defaultOnReadError))
+  const excludePaths = cmdFlagValueToArray(cli.flags['excludePaths'])
+
+  await handleManifestSetup(
+    cwd,
+    Boolean(defaultOnReadError),
+    Boolean(dynamicSbomInference),
+    excludePaths,
+  )
 }
--- a/src/commands/manifest/cmd-manifest-setup.mts
+++ b/src/commands/manifest/cmd-manifest-setup.mts
@@ -37,7 +37,7 @@ const config: CliCommandConfig = {
       type: 'boolean',
       hidden: true,
       description:
-        'After configuring CWD, recursively discover every gradle/sbt/maven build root beneath it and mark `disabled: true` on whatever matches --exclude-paths; everything else is left untouched',
+        'After configuring CWD, recursively discover every gradle/sbt/maven build root beneath it. A build root matching --exclude-paths is bulk-disabled with no prompt; every other one gets an interactive configure-or-inherit-defaults prompt',
     },
   },
   help: (command, config) => `
--- a/src/commands/manifest/cmd-manifest-setup.mts
+++ b/src/commands/manifest/cmd-manifest-setup.mts
@@ -37,7 +37,7 @@ const config: CliCommandConfig = {
       type: 'boolean',
       hidden: true,
       description:
-        'After configuring CWD, recursively discover every gradle/sbt/maven build root beneath it. A build root matching --exclude-paths is bulk-disabled with no prompt; every other one gets an interactive configure-or-inherit-defaults prompt',
+        'Recursively scans for every gradle/sbt/maven build root beneath CWD first, so the CWD config step only asks about ecosystems actually found somewhere in the tree. A build root matching --exclude-paths is bulk-disabled with no prompt, applied unconditionally; eligible build roots found afterward can be configured individually',
     },
   },
   help: (command, config) => `
--- a/src/commands/manifest/cmd-manifest-setup.mts
+++ b/src/commands/manifest/cmd-manifest-setup.mts
@@ -8,6 +8,7 @@ import { commonFlags } from '../../flags.mts'
 import { cmdFlagValueToArray } from '../../utils/cmd.mts'
 import { meowOrExit } from '../../utils/meow-with-subcommands.mts'
 import { getFlagListOutput } from '../../utils/output-formatting.mts'
+import { assertValidExcludePaths } from '../scan/exclude-paths.mts'
 
 import type {
   CliCommandConfig,
@@ -105,6 +106,7 @@ async function run(
   }
 
   const excludePaths = cmdFlagValueToArray(cli.flags['excludePaths'])
+  assertValidExcludePaths(excludePaths)
 
   await handleManifestSetup(
     cwd,
--- a/src/commands/manifest/discover-manifest-roots.mts
+++ b/src/commands/manifest/discover-manifest-roots.mts
@@ -32,7 +32,7 @@ for (const tool of BUILD_TOOLS) {
   }
 }
 
-async function realpathOrResolved(dir: string): Promise<string> {
+export async function realpathOrResolved(dir: string): Promise<string> {
   try {
     return await fs.realpath(dir)
   } catch {
--- a/src/commands/manifest/discover-manifest-roots.mts
+++ b/src/commands/manifest/discover-manifest-roots.mts
@@ -32,6 +32,29 @@ for (const tool of BUILD_TOOLS) {
   }
 }
 
+// A disabled ecosystem is skipped entirely by the glob scan below, so a
+// nested socket.json re-enabling it would never even be searched for. Callers
+// that need the cascade (not just the root file) to decide skip vs. include
+// must strip `disabled` here first, and let their own per-directory cascade
+// check apply it instead.
+export function withoutDisabledFlags(sockJson: SocketJson): SocketJson {
+  const manifest = sockJson.defaults?.manifest
+  if (!manifest) {
+    return sockJson
+  }
+  const stripped: Record<string, unknown> = { ...manifest }
+  for (const tool of BUILD_TOOLS) {
+    const section = manifest[tool]
+    if (section?.disabled) {
+      stripped[tool] = { ...section, disabled: false }
+    }
+  }
+  return {
+    ...sockJson,
+    defaults: { ...sockJson.defaults, manifest: stripped },
+  } as SocketJson
+}
+
 export async function realpathOrResolved(dir: string): Promise<string> {
   try {
     return await fs.realpath(dir)
--- a/src/commands/manifest/discover-manifest-roots.mts
+++ b/src/commands/manifest/discover-manifest-roots.mts
@@ -53,13 +53,9 @@ function sortByDepthThenPath(dirs: readonly string[], cwd: string): string[] {
   })
 }
 
-// Recursively discovers gradle/sbt/maven build-tool roots under `cwd`, one
-// filesystem pass for all three ecosystems' marker files (fast-glob unions the
-// patterns internally). Each ecosystem's candidate list is depth-sorted
-// (root-most first) so a reactor/multi-project root is always visited before
-// its own members; the caller uses that ordering plus the resulting facts
-// SBOM's `projects[].subprojectDir` to avoid re-invoking the manifest script on
-// directories a parent build root already covers.
+// Depth-sorted (root-most first) so the caller, using each build root's
+// `subprojectDir` facts, visits a reactor root before its own members and can
+// skip subprojects a parent root already covers.
 export async function findBuildToolCandidates({
   cwd,
   excludePaths,
--- a/src/commands/manifest/discover-manifest-roots.mts
+++ b/src/commands/manifest/discover-manifest-roots.mts
@@ -1,0 +1,115 @@
+import { promises as fs } from 'node:fs'
+import path from 'node:path'
+
+import { globWithGitIgnore } from '../../utils/glob.mts'
+import { excludePathToScanIgnores } from '../scan/exclude-paths.mts'
+
+import type { BuildTool } from './scripts/build-tool.mts'
+import type { SocketJson } from '../../utils/socket-json.mts'
+
+const BUILD_TOOLS: BuildTool[] = ['gradle', 'maven', 'sbt']
+
+// One glob-suffix set per ecosystem's build-descriptor marker(s); `settings.gradle(.kts)`
+// covers Kotlin-DSL roots that have no root `build.gradle`.
+const MARKERS_BY_TOOL: Record<BuildTool, string[]> = {
+  __proto__: null,
+  gradle: [
+    'build.gradle',
+    'build.gradle.kts',
+    'settings.gradle',
+    'settings.gradle.kts',
+  ],
+  maven: ['pom.xml'],
+  sbt: ['build.sbt'],
+} as unknown as Record<BuildTool, string[]>
+
+const TOOL_BY_MARKER: Record<string, BuildTool> = {
+  __proto__: null,
+} as unknown as Record<string, BuildTool>
+for (const tool of BUILD_TOOLS) {
+  for (const marker of MARKERS_BY_TOOL[tool]) {
+    TOOL_BY_MARKER[marker] = tool
+  }
+}
+
+async function realpathOrResolved(dir: string): Promise<string> {
+  try {
+    return await fs.realpath(dir)
+  } catch {
+    return path.resolve(dir)
+  }
+}
+
+function sortByDepthThenPath(dirs: readonly string[], cwd: string): string[] {
+  return [...dirs].sort((a, b) => {
+    const relA = path.relative(cwd, a)
+    const relB = path.relative(cwd, b)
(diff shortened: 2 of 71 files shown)


src/commands/manifest/discover-manifest-roots.mts
```diff
@@ -0,0 +1,111 @@
...
+  excludePaths,
+  sockJson,
+}: {
+  cwd: string
+  excludePaths?: string[] | undefined
+  sockJson: SocketJson
+}): Promise<Map<BuildTool, string[]>> {
+  const enabledTools = BUILD_TOOLS.filter(
+    tool => !sockJson.defaults?.manifest?.[tool]?.disabled,
+  )
```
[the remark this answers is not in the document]
jfblaa: Confirmed and fixed. `findBuildToolCandidates` was gating its glob scan on the root socket.json's own `disabled` flag, so a nested socket.json re-enabling that ecosystem was never even discovered — same bug the setup wizard already had to work around locally (`withoutDisabledFlags`). Moved that helper to `discover-manifest-roots.mts` (exported) and applied it in `generateRecursiveManifests` too, so the scan always runs and the existing per-directory cascade check (which already handles re-enabling correctly) is the sole arbiter of skip vs. include. Added a regression test in `generate-recursive-manifests.test.mts` covering a root-level disable with a nested re-enable.

src/commands/manifest/run-manifest-facts.mts
```diff
@@ -52,14 +60,28 @@ export async function runManifestFacts({
...
+    if (expanded.missing) {
+      process.exitCode = 1
+      logger.fail(
+        `javaHome (\`${javaHome}\`) references \`${expanded.missing}\`, which is not set in this environment.`,
+      )
+      return
+    }
+    resolvedJavaHome = expanded.value
+  }
+
```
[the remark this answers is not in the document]
jfblaa: Confirmed and fixed. Switched this line from `logger.log` (stdout) to `logger.info` (stderr), matching every other status line in this function. This is a straightforward per-project convention here — `logger.log` is reserved for the payload a command actually prints (JSON/markdown/table output), everything else (progress, warnings, success messages) goes through `logger.info`/`warn`/`fail`/`success`, which route to stderr.

src/commands/manifest/cmd-manifest-setup.mts
```diff
@@ -89,5 +104,12 @@ async function run(
...
 
-  await handleManifestSetup(cwd, Boolean(defaultOnReadError))
+  const excludePaths = cmdFlagValueToArray(cli.flags['excludePaths'])
+
+  await handleManifestSetup(
+    cwd,
+    Boolean(defaultOnReadError),
+    Boolean(dynamicSbomInference),
+    excludePaths,
+  )
```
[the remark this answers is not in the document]
jfblaa: Confirmed and fixed. Added the same `assertValidExcludePaths(excludePaths)` call that every sibling manifest command (gradle/maven/kotlin/scala/auto/dynamic-sbom-inference) already has — this command was simply missing it.

src/commands/manifest/generate-recursive-manifests.mts
```diff
@@ -0,0 +1,245 @@
...
+
+      if (!result) {
+        const failed = Boolean(
+          process.exitCode && process.exitCode !== beforeExitCode,
+        )
+        outcomes.push({
+          dir,
+          ecosystem,
+          status: failed ? 'failed' : 'empty',
+        })
```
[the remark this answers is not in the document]
jfblaa: Confirmed and fixed. Gave `runManifestFacts` an explicit `null` (real failure) vs. `undefined` (genuinely nothing to resolve, not a failure) return, instead of the caller inferring which happened from whether `process.exitCode` changed during the call. Added a regression test that pre-sets `process.exitCode = 1` before a real failure occurs and asserts it's still correctly classified as `failed`, not `empty`.

src/commands/manifest/enumerate-workspaces.mts
```diff
@@ -0,0 +1,79 @@
...
+
+  let result
+  try {
+    result = await enumerateWorkspacesScript(ecosystem, scriptOpts)
+  } catch (e) {
+    process.exitCode = 1
+    logger.fail(
+      `Could not run the ${ecosystem} build tool` +
+        (verbose ? `: ${e}` : ' (run with --verbose for details).'),
+    )
```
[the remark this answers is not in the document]
jfblaa: Fixed — switched to the existing `getErrorMessageOr` helper from `src/utils/errors.mts` (this repo's local equivalent; `@socketsecurity/lib/errors` doesn't exist in the installed `@socketsecurity/registry` version here). Also applied the same fix to the identical pattern in `run-manifest-facts.mts` for consistency, since it had the same raw ${e} coercion.

src/commands/manifest/generate-recursive-manifests.mts
```diff
@@ -0,0 +1,244 @@
...
+    for (const dir of dirs) {
+      if (covered.has(dir)) {
+        outcomes.push({ dir, ecosystem, status: 'skippedCovered' })
+        continue
+      }
+
+      const nearestRoot = nearestDisabledRoot(dir, disabledRoots)
+      const sockJson = nearestRoot
+        ? readSocketJsonCascade(dir, nearestRoot.dir, nearestRoot.sockJson)
+        : readSocketJsonCascade(dir, cwd, rootSockJson)
```
mtorp: Candidate dirs here are realpath-resolved (`findBuildToolCandidates` → `realpathOrResolved`), but `cwd` arrives raw from the command's `path.resolve(process.cwd(), cwd)`. Pairing the two breaks in two places whenever the cwd path contains a symlink — macOS `/tmp` → `/private/tmp` and `/var/folders` → `/private/var/folders`, symlinked home/workspace dirs, several CI layouts.

**The cascade never reaches its boundary.** `readSocketJsonCascade`'s loop is:

```js
const boundary = path.resolve(boundaryDir)   // /tmp/repo
let current = path.resolve(dir)              // /private/tmp/repo/moduleA
while (current !== boundary) { /* read socket.json, walk up */ }
```

`current` can never equal `boundary`, so it walks all the way to `/`, merging every `socket.json` above the intended recursion root — including `$HOME/socket.json`. That's exactly what `socket-json.test.mts`'s "does not walk past the boundary" case asserts against; it passes only because it derives `dir` and `boundaryDir` from the same `mkdtemp` result.

**`--exclude-paths` is silently dropped** (line 195). `projectIgnorePathsToReachExcludePaths` computes `path.relative(cwd, target)` → `../../private/tmp/repo/moduleA`. No user pattern starts with that prefix, so `pathRelativeToTarget` returns `undefined` for all of them. Discovery still excludes correctly (fast-glob ignores are cwd-relative), but the build tool is never told to skip excluded subprojects — so the "exclude a broken reactor member so it doesn't abort the walk" workflow this PR added quietly stops working.

The fix is the one `setupRecursiveManifestConfig` already applies — its `discoverBuildRoots` comment even names this hazard ("otherwise macOS's /tmp -> /private/tmp breaks the comparison"); it just wasn't carried over here. Resolve once near line 147 and use it for the boundary and the exclude re-anchoring:

```js
const realCwd = await realpathOrResolved(cwd)
```

While you're in here: `covered.add(path.resolve(dir, project.subprojectDir))` (line 228) isn't realpath-resolved either, so a symlinked reactor member escapes coverage and gets re-invoked as an independent root. Same at `setup-recursive-manifest-config.mts:249`.

Worth noting that `generate-recursive-manifests.test.mts:11` mocks `readSocketJsonCascade` to return its `fallback`, which is why the suite can't see any of this — one case exercising the real function against a fixture tree would close the gap.

src/commands/manifest/enumerate-workspaces.mts
```diff
@@ -0,0 +1,82 @@
...
+    return
+  }
+
+  if (result.code !== 0 && !result.projects.length) {
+    process.exitCode = 1
+    logger.fail(
+      `The ${ecosystem} build failed (exit code ${result.code}) before producing any workspace records.`,
+    )
+    return
+  }
```
mtorp: A successful enumeration always reports at least one project — the build's own root (gradle `allprojects`, sbt `allProjectRefs`, maven's full reactor list all include it), and the root can't be excluded away since `pathRelativeToTarget` drops a pattern that equals the target. So `exit 0` with zero projects means the task never ran, and requiring `code !== 0` lets that through as success.

The concrete case is a `coana-maven-extension.jar` built before `CoanaWorkspacesLifecycleParticipant` existed: `afterProjectsRead` never fires, maven exits 0, no records file is written. `markWorkspaceCoverage` then marks only the candidate dir covered, and the wizard prompts individually for every reactor member of a build it thinks has none — the exact misclassification the fail-closed design elsewhere in this PR is meant to prevent, except silent.

Dropping the exit-code condition makes it an error, and lets the message cover both shapes:

```suggestion
  if (!result.projects.length) {
    process.exitCode = 1
    logger.fail(
      `The ${ecosystem} build produced no workspace records (exit code ${result.code}); every build reports at least its own root project, so the enumeration task did not run.`,
    )
    return
  }
```

mtorp (approved):
Nice work — the hard parts here (reactor coverage classification off `projects[].subprojectDir`, the field-level cascade, the explicit `null`-clear sentinel) are well reasoned, and the commit history shows real self-correction along the way. The comments explain *why* rather than *what*, which made this a pleasant read.

Approving, with two things I'd like fixed before this goes out — both small and localized, left as inline comments:

1. `generateRecursiveManifests` pairs realpath-resolved candidate dirs with a raw `cwd`, which silently breaks the cascade boundary and drops `--exclude-paths`. `setupRecursiveManifestConfig` already guards against exactly this; the generation path just didn't get the same treatment.
2. `enumerateWorkspaces` accepts "exit 0, zero projects" as success, which turns a stale Maven extension jar into silent degradation rather than an error.

Non-blocking observations, for whenever you get to them:

- **`break ecosystems` is broader than its rationale.** `covered` is allocated per-ecosystem, so a gradle failure carries no information about maven coverage classification — yet one failure stops every ecosystem, and the output can't tell the user that maven/sbt were never attempted. Breaking only the inner loop plus an `aborted` outcome would say so.
- **The name is already taken.** `scan create` / `scan reach` have carried a hidden `--dynamic-sbom-inference` since #1451, which passes `--maven-use-only-root-socket-facts` (use *only* root facts) and implies `--auto-manifest`. This command generates *per-root* facts recursively, via a code path the scan flow never invokes. Two unrelated features under one name, one of them now user-visible — worth settling before it's unhidden.
- **The new JVM producers have no automated coverage.** `socket-workspaces.init.gradle`, `socket-workspaces.plugin.scala`, and `SocketWorkspacesRecordsEngine` aren't referenced by `scripts/test/run-compat.sh`, which is the only thing exercising the bundled producers — and per its own README there's no CI for it. The files carry explicit compatibility claims (Gradle 1.0+, Scala 2.10/sbt 0.13 through 2.12/sbt 1.x) that nothing verifies, and since `WORKSPACES_TASK` is the sole entry point, a registration or compile error surfaces only as the silent degradation in comment 2.
- **`configureCandidate` materializes the inherited section into the child** (pinned by `setup-recursive-manifest-config.test.mts:441` — `bin: './gradlew'` comes only from the cascade and still gets written). So once a user picks "Configure", later edits to the parent's `javaHome`/`bin`/`excludeConfigs` stop reaching that candidate. Deliberate and tested, but it converts inheritance into a snapshot, which is surprising when a field-level cascade is the headline feature — worth either writing only genuinely-changed fields or saying so in the comment.
- **`--json --verbose` emits invalid JSON.** `cmd-manifest-dynamic-sbom-inference.mts:92-96` and `run-manifest-facts.mts:166,175` use `logger.log` (stdout) ahead of the payload. This is the only manifest command carrying `outputFlags`, so the combination is newly reachable — and 58fa321 fixed exactly this one line away.
- **Smaller stuff:** two stale `socket manifest setup --recursive` references (`enumerate-workspaces.mts:15`, `SocketWorkspacesRecordsEngine.java:17` — the flag is `--dynamic-sbom-inference`); `expandEnvVarRefs` has no `$$` escape and reports only the first missing variable; `renderTable` emits a bare blank line when every outcome is `skippedCovered`; the exclude-glob variant expansion now lives in six producers, each commented as mirroring the others, so a shared fixture-driven test would keep them honest; the generation command is visible while its only config surface is hidden.

The `e2e-tests` failures on Node 20/22/24 are `socket fix` on a Python/django fixture, unrelated to this PR — type check, lint, and the unit matrix are green.

src/commands/manifest/generate-recursive-manifests.mts
jdalton: Confirmed independently, and the boundary walk is the sharper half.

`socket-json.mts:207-212` compares `path.resolve(dir)` against `path.resolve(boundaryDir)` with `!==`, and `path.resolve` does not resolve symlinks. Pairing the realpath-resolved `dir` from `findBuildToolCandidates` with the raw `cwd` at `generate-recursive-manifests.mts:169` means the two strings cannot converge on any host where the temp or home path is a symlink, so the walk runs to `/` and merges every `socket.json` above the intended root.

The `realpathOrResolved(cwd)` fix you sketched covers both the boundary and the exclude re-anchoring. The test note is worth taking too: with `readSocketJsonCascade` mocked to return its fallback, no current case can observe any of this.

src/commands/manifest/enumerate-workspaces.mts
jdalton: Confirmed. `enumerate-workspaces.mts:73` reads `result.code !== 0 && !result.projects.length`, so an exit-0 run with an empty project list falls straight through to `return { projects: result.projects }` and is reported as success.

Since every supported build reports at least its own root, an empty list can only mean the enumeration task never ran. Dropping the exit-code condition makes that fail closed, matching the rest of the PR.

src/commands/manifest/generate-recursive-manifests.mts
jfblaa: Confirmed and fixed - great catch, this was a real bug. Resolved `cwd` once via `realpathOrResolved` (mirroring what `setupRecursiveManifestConfig` already does) and used that for both `readSocketJsonCascade`'s boundary and `projectIgnorePathsToReachExcludePaths`'s `cwd`. Also realpath-resolved each `project.subprojectDir` before adding it to `covered`, here and at the `setup-recursive-manifest-config.mts:249` spot you flagged, so a symlinked reactor member is recognized as covered instead of escaping and being reinvoked as independent.

Added a regression test that creates a real symlinked cwd + an ancestor socket.json strictly above the intended boundary and asserts that ancestor's setting never leaks in - confirmed it fails against the old code (walks past the boundary and picks up the ancestor's `bin`) and passes with the fix. Didn't restructure the existing mocked tests to exercise the real cascade function since this new one covers the actual gap directly.

src/commands/manifest/enumerate-workspaces.mts
jfblaa: Confirmed and fixed, using your suggested change. Dropped the `result.code !== 0` condition so any run producing zero project records fails, regardless of exit code - matches the invariant that a real enumeration always reports at least its own root project. Added a test covering the exact case you called out (clean exit, zero projects) to make sure a stale/pre-participant extension jar can't silently look like an empty-but-successful run.

jfblaa:
Thanks for the thorough review! Addressed the non-blocking observations:

- **Naming collision** — made `socket manifest dynamic-sbom-inference` hidden too, so nothing user-visible collides with the existing `scan create`/`scan reach` `--dynamic-sbom-inference` flag while the two get sorted out. Noted the collision in a comment so it isn't forgotten.
- **`configureCandidate` materializes the inherited section** — fixed. It now diffs the sub-wizard's final result against an ancestor-only baseline (the cascade from `dir`'s parent, excluding `dir`'s own file) and only writes fields that actually differ, so an untouched field keeps inheriting live instead of getting pinned. Updated the test that was asserting the old pin-everything behavior.
- **New JVM producers had no automated coverage** — added `smoke-test-workspaces.sh` for each ecosystem, wired into the same per-version matrix `run-compat.sh` already runs for the facts scripts. Verified locally against the installed gradle/maven/sbt (didn't run the full multi-version matrix, but the wiring and assertions are in place and passing).
- **`--json --verbose` emitted invalid JSON** — gated the verbose debug preamble on `!json` and switched `run-manifest-facts.mts`'s verbose detail logging to stderr.
- Fixed the two stale `--recursive` comments and the `renderTable` bare-blank-line case.

Holding off on **`break ecosystems` being broader than its rationale** — want to talk through the tradeoff (a maven failure currently also aborts sbt/gradle, even though coverage is tracked per-ecosystem) before changing the fail-closed scope, given how much iteration went into establishing that behavior. Will follow up separately.

Left `expandEnvVarRefs`'s `$$`-escape/first-missing-var reporting and the shared fixture-driven test for the exclude-glob expansion (duplicated across the 6 producers) as-is for now — lower priority than the above, happy to pick those up too if you'd rather not defer them.

jfblaa:
Follow-up on the two items from my earlier comment:

- **`break ecosystems` scope** — talked it through and implemented it. The abort now only applies to the failing ecosystem's own remaining walk (`covered`/`disabledRoots` are per-ecosystem, so a maven failure has nothing to say about gradle/sbt's classification). Added an `'aborted'` outcome status for that ecosystem's still-untried candidates, so they show up explicitly in the output instead of silently vanishing. An unrelated ecosystem at the same directory now runs to completion as expected.
- **`expandEnvVarRefs`** — did both: it now reports every missing variable in one pass (not just the first), and `$$` expands to a literal `$` so a value can contain a literal `$WORD` without being treated as a reference. Added a dedicated test file, since there wasn't one.

Left the shared fixture-driven test for the exclude-glob expansion (duplicated across the 6 producers) deferred — didn't want to expand scope further on this PR, but happy to take it on as a follow-up if you'd like.

jfblaa:
Followed up on the "exclude-glob variant expansion now lives in six producers" observation: centralized `--exclude-paths` glob compilation into a single TypeScript module (`exclude-paths-glob.mts`), mirroring the pattern from #1404's `config-glob.mts`. Each of the six gradle/maven/sbt facts/workspaces producers now just `Pattern.compile()`s a pre-compiled regex pattern source handed to it via the existing `-D`/`-P` property mechanism, instead of carrying its own NIO-PathMatcher-based glob implementation and zero-depth-variant workaround.

Gradle/maven/sbt all run on the JVM and share `java.util.regex`, so a single regex dialect covers every producer (simpler than #1404's three-way JS/Java/.NET contract). Added a 26-case cross-language vector test suite as the single source of truth for matching semantics, and verified end-to-end against real gradle, maven, and sbt invocations (both facts and workspaces variants) with an actual `--exclude-paths` value excluding a subproject.

PR #1484, opened 2026-08-06
socket-cli: multi-root support for --dynamic-sbom-inference

jfblaa:
## Summary
- `socket scan create --dynamic-sbom-inference` (with or without `--reach`) now generates Socket facts recursively across every independent gradle/sbt/maven build root under the scan target, instead of only the one at the scan root.
- Redesigned the resolved-artifact-paths sidecar passed to Coana (`--compute-artifacts-sidecar`) to be keyed by build root rather than a single flat, cross-root-deduplicated list, so two independent build roots emitting modules with the same package identity can never collide.
- sbt builds discovered across multiple independent roots in one run now share a single toolchain-provisioning directory instead of re-provisioning it per root (verified empirically: the plugin and its Scala compiler-bridge compile exactly once per process).
- Bumped `@coana-tech/cli` to `15.10.5`, the published version with matching consumer-side support for the new sidecar format.

Paired with the Coana-side consumer work (REA-702) — both were needed together and have been verified working end-to-end, including against the published `@coana-tech/cli@15.10.5` (not just a local dev build).

Linear: REA-704

## Test plan
- [x] `pnpm run check:tsc` (both projects)
- [x] `pnpm run lint --all`
- [x] Full `src/commands/manifest` + `src/commands/scan` + `src/utils` + `src/commands/manifest/scripts` test suites (686 tests)
- [x] Real end-to-end run against a multi-root sandbox repo (independent Maven reactor, Gradle multi-project, standalone Gradle lib nested inside the Maven reactor, sbt project), using the published `@coana-tech/cli@15.10.5` — all expected reachable/unreachable results present per project

---

> [!NOTE]
> **Medium Risk**
> Changes the reachability auto-manifest path and the frozen Coana sidecar contract; incorrect behavior would skew multi-root JVM scans, but the bump is paired with Coana 15.10.5 and coverage is extensive in manifest/scan tests.
> 
> **Overview**
> **`--dynamic-sbom-inference`** (with `--auto-manifest`) now recursively generates `.socket.facts.json` for every independent Gradle, Maven, and sbt root under the scan target, not only the cwd. JVM generation in `generateAutoManifest` is skipped for those ecosystems when this flag is on; conda/bazel still use auto-manifest. Generated facts paths are merged into scan targets.
> 
> The **resolved-artifact sidecar** for Coana (`--compute-artifacts-sidecar`) is redesigned: keys are absolute `.socket.facts.json` paths, each holding that file’s `projects` and `components` with attached `targets`/`sources` (full SBOM-shaped entries). Cross-reactor coordinate deduplication is removed so identical module names in different reactors cannot collide. Recursive and auto-manifest sidecars are merged when both run; reachability uses **`--maven-use-only-socket-facts`** instead of **`--maven-use-only-root-socket-facts`**.
> 
> **sbt** roots in one run share a single temp global base (Scala toolchain cache) when multiple sbt projects are discovered. Sidecar keys use symlink-resolved facts paths via shared **`realpathOrResolved`** in `utils/fs.mts`.
> 
> Bumps **`@coana-tech/cli`** to **15.10.5** (consumer for the new sidecar shape).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b5130facdababef654ccd7c457784133af27d7f1. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/commands/manifest/scripts/sidecar.mts
```diff
@@ -1,111 +1,147 @@
...
-  classifier: string | null
-  // Classpath entries (jars / first-party output dirs).
-  targets: string[]
-  // First-party source roots; [] for external deps.
-  sources: string[]
+export type SidecarComponentEntry = SocketFactsSbomComponent & {
+  // Classpath entries (jars, or a sibling first-party project's own build
+  // output dirs when this dependency edge resolves to one - REA-687). `[]`
+  // means resolution was attempted and found nothing (e.g. a pom/BOM);
+  // undefined means resolution couldn't be attempted at all (see attachPaths).
```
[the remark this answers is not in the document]
jfblaa: Removed the tracker reference — the comment now just describes the mechanism without the ticket ID. Fixed in f1f189169.

src/commands/scan/handle-create-new-scan.mts
```diff
@@ -162,10 +202,17 @@ export async function handleCreateNewScan({
...
+          ? mergeResolvedPathsSidecars(
+              resolvedPathsSidecar,
+              autoManifestResult.resolvedPathsSidecar,
+            )
+          : autoManifestResult.resolvedPathsSidecar
+      }
       if (autoManifestResult.generatedFiles.length) {
         scanTargets = Array.from(
-          new Set([...targets, ...autoManifestResult.generatedFiles]),
+          new Set([...scanTargets, ...autoManifestResult.generatedFiles]),
```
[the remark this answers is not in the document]
jfblaa: Good catch — fixed. handleCreateNewScan now checks outcomes for any status === 'failed' and throws an InputError before scanTargets/the sidecar are ever touched, matching handleManifestDynamicSbomInference's own check, so a partial multi-root run aborts loudly instead of uploading/running reach on an incomplete facts set. Added a regression test asserting getPackageFilesForScan and fetchCreateOrgFullScan are never called when a root fails. Fixed in f1f189169.

src/commands/manifest/generate-recursive-manifests.mts
```diff
@@ -252,10 +202,117 @@ export async function generateRecursiveManifests({
...
+    })
+
+  // A shared global base across every sbt root in this run lets sbt's own
+  // Scala-toolchain cache under <base>/boot survive between invocations
+  // instead of being reprovisioned per root (the plugin file is rewritten and
+  // records.tsv is fully overwritten - not appended - on every invocation, so
+  // reuse is safe). Skipped entirely when there's no sbt root to benefit.
+  const outcomes = candidatesByTool.get('sbt')?.length
+    ? await withTmpDir('socket-sbt-facts-shared-', runAll)
+    : await runAll(undefined)
```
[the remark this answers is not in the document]
jfblaa: Confirmed and fixed. generateRecursiveManifests now accepts an optional caller-supplied sbtTmpDir — when given, it's reused as-is for every sbt root and never cleaned up by generateRecursiveManifests itself (the caller owns that lifetime), exactly mirroring how ManifestScriptOptions.tmpDir already works one level down. handleCreateNewScan passes its existing manifestTmpDir (already kept alive until reachability analysis finishes, per the comment right above the withTmpDir call) whenever reach.runReachabilityAnalysis is true, so sbt's shared global base (and the Scala toolchain paths under its boot dir) now survives through reach instead of being deleted the moment recursive generation returns. Added a test proving a caller-supplied sbtTmpDir is used for every sbt invocation and is still on disk afterward. Fixed in f1f189169.

src/commands/scan/reachability-flags.mts
```diff
@@ -9,7 +9,7 @@ export const reachabilityFlags: MeowFlags = {
     default: false,
     hidden: true,
     description:
-      'Internal: enables dynamic SBOM inference for full application reachability analysis. Passes --maven-use-only-root-socket-facts to Coana and implies --auto-manifest.',
+      'Internal: enables dynamic SBOM inference for full application reachability analysis. Recursively generates Socket facts for every independent gradle/sbt/maven build root, passes --maven-use-only-socket-facts to Coana, and implies --auto-manifest.',
```
mtorp: Do we feel confident enough in this feature to remove the `internal` note and the `hidden` flag? If yes, we should also update the docs accordingly.

mtorp (approved):
LGTM ✅

src/commands/scan/reachability-flags.mts
jfblaa: Keeping it hidden/internal for now — this is the first release with multi-root support, and there's a followup fix landing in #1485 (an escaping-subprojectDir coverage bug in the recursive discovery) plus REA-705's further subprojectPath/workspacePath generalization still in progress on the Coana side, which may warrant another `@coana-tech/cli` bump. Want to get more field validation under our belt before committing to public docs and a stable flag surface.

jfblaa: Update: went ahead and un-hid it (commit 07d0f36c5) — the Coana side has now shipped, so bumped to `@coana-tech/cli` 15.10.8 and dropped the `hidden`/'Internal:' status. Also clarified the --help description: for Gradle, sbt, and Maven, only dependencies present in the generated Socket facts SBOM files are used for analysis; other ecosystems are unaffected. Added a CHANGELOG entry too.

src/commands/manifest/cmd-manifest-setup.test.mts
```diff
@@ -24,6 +24,8 @@ describe('socket manifest setup', async () => {
 
           Options
             --default-on-read-error  If reading the socket.json fails, just use a default config? Warning: This might override the existing json file!
+            --dynamic-sbom-inference  Recursively scans for every gradle/sbt/maven build root beneath CWD first, so the CWD config step only asks about ecosystems actually found somewhere in the tree. A build root matching --exclude-paths is bulk-disabled with no prompt, applied unconditionally; eligible build roots found afterward can be configured individually
```
mtorp: This is a pretty long description. Maybe it should talk more about the focus on generating the SBOMs using the package managers? Something like 'Recursively generates dynamic SBOMs using the package manager tools for SBT, Maven and Gradle' Maybe even include a sentence about this resulting in more accurate SBOMs for ecosystems and package managers where static resolution is hard.

src/commands/scan/reachability-flags.mts
```diff
@@ -7,9 +7,8 @@ export const reachabilityFlags: MeowFlags = {
   dynamicSbomInference: {
     type: 'boolean',
     default: false,
-    hidden: true,
     description:
-      'Internal: enables dynamic SBOM inference for full application reachability analysis. Passes --maven-use-only-root-socket-facts to Coana and implies --auto-manifest.',
+      'Enables per-project, per-subproject/module/workspace reachability analysis for Gradle, sbt, and Maven, instead of the default coarse-grained analysis where everything is treated as one synthetic root. Recursively discovers every independent build root and generates a Socket facts SBOM for each (implies --auto-manifest). Applies to reachability analysis only: for Gradle, sbt, and Maven, only dependencies present in these generated Socket facts SBOM files are considered reachable; the scan itself still detects all dependencies as normal, and other ecosystems are unaffected.',
```
mtorp: Maybe include a short sentence about what a 'Socket Facts SBOM', e.g., something like 'an SBOM inferred by the package manager tools'.

mtorp (approved):
LGTM ✅

src/commands/manifest/cmd-manifest-setup.test.mts
jfblaa: Trimmed it down considerably (commit 2265185cf): "Generates dynamic SBOMs via the Gradle/sbt/Maven package manager tools for more accurate results than static resolution. Scans every build root under CWD first, so the configurator only asks about ecosystems actually found; --exclude-paths matches are bulk-disabled, others configured individually."

src/commands/scan/reachability-flags.mts
jfblaa: Reworked and shortened (commit 2265185cf), now explains what a Socket facts SBOM is: "For Gradle, sbt, and Maven: splits reachability analysis per project/module using a Socket facts SBOM (generated directly by each package manager) per build root, instead of one synthetic root. Reachability analysis only; implies --auto-manifest."

PR #1493, opened 2026-08-13
fix(scan): reject every --reach-* modifier when --reach is not enabled

mtorp:
## Problem

Running `socket scan create --reach-debug .` completes successfully as a plain SBOM scan — no reachability analysis runs, and there is no warning that the flag did nothing. A user can easily believe they have a working reachability scan and ship that report as a reachability assessment.

`socket scan create` already has a guard ("Reachability analysis flags require --reach to be enabled"), but it relied on a hand-maintained list of flags that had drifted out of sync. Four boolean flags were missing and slipped through silently:

- `--reach-debug`
- `--reach-detailed-analysis-log-file`
- `--reach-disable-external-tool-checks`
- `--reach-retain-facts-file`

## Fix

Derive the boolean-flag portion of the guard from the `reachabilityFlags` definition itself (comparing each boolean flag against its declared default), so newly added reach flags are covered automatically instead of relying on the list being kept up to date. The deprecated no-op `--reach-disable-analysis-splitting` remains exempt on purpose.

With this change, `socket scan create --reach-debug .` exits non-zero with:

```
✖ Reachability analysis flags require --reach to be enabled (add --reach flag to use --reach-* options)
```

## Tests

- New: `--reach-debug` without `--reach` fails.
- New: `--reach-retain-facts-file` without `--reach` fails.
- New: deprecated no-op `--reach-disable-analysis-splitting` without `--reach` still succeeds.
- All 33 `cmd-scan-create` tests pass; `pnpm run check` passes.

### DIFF
--- a/src/commands/scan/cmd-scan-create.mts
+++ b/src/commands/scan/cmd-scan-create.mts
@@ -504,30 +504,28 @@ async function run(
   const isUsingNonDefaultConcurrency =
     reachConcurrency !== reachabilityFlags['reachConcurrency']?.default
 
-  const isUsingNonDefaultAnalytics =
-    reachDisableAnalytics !==
-    reachabilityFlags['reachDisableAnalytics']?.default
-
   const isUsingNonDefaultVersion =
     reachVersion !== reachabilityFlags['reachVersion']?.default
 
+  // Compare every boolean reach flag against its declared default so newly
+  // added flags require --reach automatically instead of relying on a
+  // hand-maintained list. The deprecated no-op
+  // --reach-disable-analysis-splitting is excluded on purpose.
+  const isUsingAnyBooleanReachFlag = Object.entries(reachabilityFlags).some(
+    ({ 0: name, 1: flag }) =>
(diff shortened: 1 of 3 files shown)


src/commands/scan/cmd-scan-create.mts
```diff
@@ -504,30 +504,28 @@ async function run(
...
-
   const isUsingNonDefaultVersion =
     reachVersion !== reachabilityFlags['reachVersion']?.default
 
+  // Compare every boolean reach flag against its declared default so newly
+  // added flags require --reach automatically instead of relying on a
+  // hand-maintained list. The deprecated no-op
+  // --reach-disable-analysis-splitting is excluded on purpose.
+  const isUsingAnyBooleanReachFlag = Object.entries(reachabilityFlags).some(
+    ({ 0: name, 1: flag }) =>
```
BarrensZeppelin:
```suggestion
    ([name, flag]) =>
```

BarrensZeppelin (approved):
Nice 🙂

PR #1511, opened 2026-08-24
fix(release): don't require version field to start a line

jdalton:
LLM Description written by Claude Code:Claude Sonnet 5
-----
## Summary

`writeManifestVersion` in `scripts/release/bump.mts` rewrites `package.json`'s version field with `/^(\s*"version":\s*")[^"]*(")/m` - anchored to require the field to open its own line. That anchor is fragile: the v1.x publish workflow hit the mismatch on CI twice ([2026-08-18](https://github.com/SocketDev/socket-cli/actions/runs/32138553834/job/95715503634), [2026-08-24](https://github.com/SocketDev/socket-cli/actions/runs/32710321835)), aborting the release with a false "no top-level version line" error even though `JSON.parse` on the exact same content resolves `version` correctly every time (confirmed by pulling the committed `package.json` at each failing run's exact commit SHA and testing the regex against it directly - it matches fine in isolation, so the anchor is failing against *something* in the CI-checked-out bytes that a plain `git show` doesn't reproduce).

The fleet's own canonical `scripts/fleet/bump.mts` (used by `main` and every other fleet member) already solves this the same way this PR does: no `^` anchor, no `/m` flag, just match the first `"version": "..."` occurrence anywhere in the file. `main` and other fleet repos were never at risk - only `v1.x`, which predates the fleet migration and still carries its own bespoke copy of this script.

## Changes

- `scripts/release/bump.mts`: drop the anchor/multiline flag, matching the fleet-canonical pattern
- `test/release-bump.test.mts`: new coverage for `writeManifestVersion`, including a minified-manifest case that reproduces the exact class of failure the old anchor was vulnerable to

## Test plan

- [x] `pnpm run test:unit test/release-bump.test.mts` - 3/3 pass
- [x] `pnpm run test:unit test/release-version.test.mts` - 29/29 pass (sibling suite unaffected)
- [x] `pnpm exec oxlint` on touched files - 0 warnings/errors

---

> [!NOTE]
> **Low Risk**
> Narrow change to release-time string replacement plus tests; no auth, runtime CLI, or user data paths.
> 
> **Overview**
> **Fixes intermittent CI release failures** where `writeManifestVersion` falsely reported it could not find a top-level `version` in `package.json` even though `JSON.parse` succeeded.
> 
> The bump script now replaces the version with an **unanchored** regex that matches the first `"version": "…"` anywhere in the file, instead of requiring that pattern at the start of a line (`^` + `/m`). That aligns with the fleet bump approach and handles **minified or oddly formatted** manifests that still parse as valid JSON.
> 
> Adds **`test/release-bump.test.mts`** with unit tests for pretty-printed and minified `package.json`, plus the error path when `version` is missing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 09f7713c7b78e9d73bb9f7dfad57906d5d0e4aac. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/scripts/release/bump.mts
+++ b/scripts/release/bump.mts
@@ -158,10 +158,7 @@ function readPackageJson(): { parsed: PackageJsonShape; raw: string } {
  * is the one line a reviewer expects.
  */
 export function writeManifestVersion(raw: string, version: string): string {
-  const replaced = raw.replace(
-    /^(\s*"version":\s*")[^"]*(")/m,
-    `$1${version}$2`,
-  )
+  const replaced = raw.replace(/("version":\s*")[^"]*(")/, `$1${version}$2`)
   if (replaced === raw) {
     throw new Error(
       '[bump] could not rewrite the package.json version field.\n' +
--- a/scripts/release/bump.mts
+++ b/scripts/release/bump.mts
@@ -158,7 +158,10 @@ function readPackageJson(): { parsed: PackageJsonShape; raw: string } {
  * is the one line a reviewer expects.
  */
 export function writeManifestVersion(raw: string, version: string): string {
-  const replaced = raw.replace(/("version":\s*")[^"]*(")/, `$1${version}$2`)
+  const replaced = raw.replace(
+    /("version":\s*")[^"]+(")/,
+    (_m, pre: string, post: string) => `${pre}${version}${post}`,
+  )
   if (replaced === raw) {
     throw new Error(
       '[bump] could not rewrite the package.json version field.\n' +
--- a/test/release-bump.test.mts
+++ b/test/release-bump.test.mts
@@ -37,4 +37,11 @@ describe('writeManifestVersion', () => {
       /could not rewrite the package\.json version field/,
     )
   })
+
+  it('throws when the version field is already empty', () => {
+    const raw = '{\n  "name": "socket",\n  "version": ""\n}\n'
+    expect(() => writeManifestVersion(raw, '1.1.160')).toThrow(
+      /could not rewrite the package\.json version field/,
+    )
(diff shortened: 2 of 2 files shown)


jdalton:
[agent] Follow-up commit brings `writeManifestVersion` to exact parity with the fleet's canonical `scripts/fleet/bump.mts` (`replaceVersion`): `[^"]+` instead of `[^"]*` (an empty version is as broken as a missing one), and a replacer function instead of a `\$1\${version}\$2` template (avoids mis-substitution if a derived version ever contained a literal `$`).

Also worth noting while double-checking this: the canonical `replaceVersion` itself has no check that the replace actually matched - it can silently no-op and ship a release with the stale version still in `package.json`. This PR's `writeManifestVersion` throws loudly instead (kept intentionally, not something I'm changing to "match" canonical). Flagging as a separate, low-urgency hardening idea for the fleet script rather than folding it into this PR.

Separately: this PR's own merge state is currently `UNSTABLE` - GitHub is waiting on `e2e-tests (20, ubuntu-latest)` as a required check, but the workflow only runs `(22)` and `(24)` now. That's exactly SURF-1618, live on this PR.

jdalton:
[agent] Correction to my last comment: this PR's `UNSTABLE` merge state was just pending checks finishing, not a stale required check.

<details>
<summary>Details</summary>

Verified directly - v1.x has no classic branch protection and its ruleset only enforces `deletion`/`non_fast_forward`, no required status checks at all. That part of SURF-1618 was already resolved before this PR (PR #1485 merged fine). Closed the ticket.
</details>

PR #1513, opened 2026-08-25
feat(scan): support --dynamic-sbom-inference on scan reach

jfblaa:
## What

Adds `--dynamic-sbom-inference` support to `socket scan reach`.

The flag was hidden and hardcoded to `false` there, because it relied on `--auto-manifest` generating the per-build-root Socket facts first and `scan reach` has no such flag. It now runs the recursive facts generation directly instead, so standalone reachability gets the same per-project/module splitting `scan create` does.

## Why

`socket scan reach` is the path Coana uses for its CI e2e tests. Without the flag, Gradle and sbt projects analyzed through it are reduced to one synthetic root and every test passes vacuously with 0 vulnerabilities detected.

## How

- **New `run-dynamic-sbom-inference.mts`** — the recursive-facts step, extracted from `handle-create-new-scan.mts` so both commands share one implementation and one set of error messages: discover every independent gradle/sbt/maven build root under `cwd`, generate its `.socket.facts.json`, accumulate the resolved-paths sidecar, and throw `InputError` on "no build root found" / "a build root failed".
- **`handle-scan-reach.mts`** — when the flag is on, runs that step before manifest discovery, merges the generated facts paths into the targets handed to `getPackageFilesForScan` (so they get uploaded), and forwards the sidecar to `performReachabilityAnalysis`, which is what makes it pass `--maven-use-only-socket-facts` alongside `--compute-artifacts-sidecar`. The body is wrapped in a `withTmpDir` that only allocates when the flag is on, since sbt's Scala toolchain lives under the shared global base the sidecar's paths point into and must outlive the analysis.
- **`cmd-scan-reach.mts`** — flag unhidden and actually read. Its description differs from `scan create`'s (no `--auto-manifest` here; the build tools are run directly), so the shared sentence moved to `DYNAMIC_SBOM_INFERENCE_DESCRIPTION` in `reachability-flags.mts` and each command appends its own tail.
- **`handle-create-new-scan.mts`** — unchanged behavior, now delegating to the shared helper.

No changelog entry: `socket scan reach` is internal-only.

## Testing

- `pnpm run check` passes.
- `pnpm test:unit src/commands/scan/ src/commands/manifest/` — 748 passed / 2 skipped, including 4 new tests for the helper and 3 for the `scan reach` wiring.
- `scan reach --help` snapshot regenerated.

## Note

The generated per-root facts are written as `<build-root>/.socket.facts.json`, and Coana's report defaults to `.socket.facts.json` in `cwd`. If `cwd` is itself a build root, the report overwrites that root's generated facts after the upload. Pre-existing behavior shared with `scan create`, not introduced here — `--output` avoids it if a caller needs both files on disk.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Changes reachability manifest discovery and Coana inputs for multi-root JVM builds; failures abort the whole command rather than partial analysis, but behavior only applies when the new flag is set.
> 
> **Overview**
> **`socket scan reach`** now honors **`--dynamic-sbom-inference`** instead of hiding the flag and forcing it off. For Gradle, sbt, and Maven monorepos, the command discovers each independent build root, builds it to emit per-root **`.socket.facts.json`**, includes those paths in manifest upload/discovery, and passes the resolved-paths sidecar into Coana so reachability splits per module like **`scan create`**.
> 
> Recursive facts generation is centralized in **`runDynamicSbomInference`**, shared with **`handle-create-new-scan`** (same errors for missing roots or partial failures). **`scan reach`** runs that step directly (no **`--auto-manifest`**); when the flag is on it keeps an **`sbt`** toolchain temp dir alive via **`withTmpDir`** until analysis finishes. Help text uses a shared **`DYNAMIC_SBOM_INFERENCE_DESCRIPTION`** with command-specific wording.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 84aafcf31c5e1e6ea4aef7c3681e4092db4d0760. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/commands/scan/cmd-scan-reach.mts
+++ b/src/commands/scan/cmd-scan-reach.mts
@@ -4,7 +4,11 @@ import { logger } from '@socketsecurity/registry/lib/logger'
 
 import { assertValidExcludePaths } from './exclude-paths.mts'
 import { handleScanReach } from './handle-scan-reach.mts'
-import { excludePathsFlag, reachabilityFlags } from './reachability-flags.mts'
+import {
+  DYNAMIC_SBOM_INFERENCE_DESCRIPTION,
+  excludePathsFlag,
+  reachabilityFlags,
+} from './reachability-flags.mts'
 import { suggestTarget } from './suggest_target.mts'
 import { validateReachabilityTarget } from './validate-reachability-target.mts'
 import constants from '../../constants.mts'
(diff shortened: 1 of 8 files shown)


mtorp (approved):
LGTM ✅

PR #1532, opened 2026-09-04
refactor(scan): require a manifests tar hash for reachability analysis

BarrensZeppelin:
## What

`performReachabilityAnalysis` guarded its manifest upload on `uploadManifests && orgSlug && packagePaths`, all three of which were optional in `ReachabilityAnalysisOptions`. Falling through that guard left `tarHash` undefined, which dropped both `--manifests-tar-hash` **and** `--run-without-docker` from the Coana invocation.

That fallback is a different execution mode, not a missing flag: Coana would attempt Docker and resolve manifests from the local tree, which is deprecated.

## Why now

No live caller can currently hit it. Of the callers that reach reachability analysis, `scan reach` and `scan create` both gate on `!!orgSlug` in their `cmd-*` files and on `packagePaths.length > 0` before calling; `scan github` and `ci` don't gate `orgSlug` but hard-code `runReachabilityAnalysis: false`. So the `else` branch was dead — but nothing in the type stopped a new caller from reviving it.

## Changes

- `orgSlug` and `packagePaths` are now required (`string` / `string[]`); the `uploadManifests` option is gone.
- The `options` parameter is required rather than `options?` — it had required fields already.
- The upload block is flattened to straight-line code, so `tarHash` is a definite `string` by the time `coanaArgs` is built and the two flags are unconditional. A hash-less upload response still fails hard at the parse boundary.

## Also: `--lazy-mode` is no longer forwarded

Coana dropped support for `--lazy-mode`, so the CLI must stop passing it. `reachLazyMode` is gone from `ReachabilityOptions` and from every call site that was threading it through.

The hidden `--reach-lazy-mode` flag itself stays accepted and is now swallowed, so existing invocations don't start erroring — the same retirement `--reach-disable-analysis-splitting` got. Its description says it's a deprecated no-op.

One knock-on: `scan create` decides "you passed a reach flag, so you need `--reach`" by comparing every boolean reach flag against its declared default, with a hand-maintained exclusion for `reachDisableAnalysisSplitting`. That exclusion becomes a `DEPRECATED_NO_OP_REACH_FLAGS` set holding both names, so `--reach-lazy-mode` on its own no longer trips the requires-`--reach` error.

## Tests

The suite previously leaned on the no-upload branch to keep the SDK/API import chains out, so `setupSdk` and `handleApiCall` were bare `vi.fn()`s returning `undefined`. They now return an SDK stub and a tar hash, and all call sites pass the required options. Two new tests: the hash and `--run-without-docker` always reach Coana, and a hash-less upload fails without spawning Coana at all.

`pnpm test:unit` on the three affected scan test files: 39 passed. `pnpm run lint`: passed.

No changelog entry — internal, no user-visible behavior change.


mtorp (approved):
Awesome thanks!

